PR #16, opened 2015-04-07
devel-pg floatingpoint

patayg:
major cleanup of floating point constants and random number usage


BishopWolf:
I have tested these changes and they are safe for all my macros

dsarrut:
Thanks a lot Gergely !

PR #17, opened 2015-04-14
Changes, cleanings and other stuffs

BishopWolf:
These changes are unobtrusive in sequential Gate, but in order to move to MT they are necessary, all dynamic calls to G4RunManager are rewritten to use GateRunManager. Also there is a little cleanup of the code. In my system Gate runs with all these changes applied.


BishopWolf:
I switch occurrences of G4endl to Gateendl because there are a mixing of both and there shall be only one for code sanity. This also serves the purpose of refactoring all G4cout occurrences to GateMessages, since it incorporates required includes in the respective files.

dsarrut:
thanks a lot Alex ! Please for the future, if possible try to propose several smaller "pull request" rather than a single big one, it makes it more simple for me to update. 

thank again,

PR #20, opened 2015-05-07
fix compatibility with geant4.10.01.p01

mojca:
This is a patch to make Gate compatible with Geant4.10.01.p01, following suggestions from Andrea Dotti.

It solves the compilation issues, but please test it carefully first.

See also http://bugzilla-geant4.kek.jp/show_bug.cgi?id=1739


dsarrut:
seems good for me, thanks !

Issue #27, opened 2015-06-02
Give the user the chance to declare actor properties the way he wants


BishopWolf:
In the file GateVImageActor in the void GateVImageActor::Construct()
you expect for the user to declare the size of the output matrix
(let's call it S), and the the user has to choose between resolution
(R) OR voxelSize (VS), being S = R * VS. This is an annoying behavior
for the user because you may know VS and R, and then you must multiply
to put in the actor the Size. So why not using any possible
combination? Solution is easy to implement.

Details

void GateVImageActor::Construct()
...
if (!mHalfSizeIsSet){
    mHalfSize = ComputeBoundingBox(mVolume->GetLogicalVolume()->GetSolid());
  }
  if (mResolutionIsSet && mVoxelSizeIsSet) {
    GateError("GateVImageActor -- Construct: Please give the
resolution OR the voxelsize (not both) for the sensor");
  }
  if (!mResolutionIsSet && !mVoxelSizeIsSet) {
    mResolution = G4ThreeVector(1.0, 1.0, 1.0);
    mResolutionIsSet = true;
  }

Can be replaced by

 if (!mHalfSizeIsSet){ //the user doesn't specify the size
          if (mResolutionIsSet && mVoxelSizeIsSet){ //but specify resolution
and voxelsize
                  mHalfSize = KroneckerProduct(mResolution, mVoxelSize)/2; //compute
size as the dot product of resolution and voxelsize
          }
          else { // also it doesn't specify one of the resolution or
voxelsize variables
                  mHalfSize =
ComputeBoundingBox(mVolume->GetLogicalVolume()->GetSolid()); // so get
size from the volume the actor is attached to
          }
          mHalfSizeIsSet = true; // say that size is set
  }
  else { // the user specify the size
          if (mResolutionIsSet && mVoxelSizeIsSet) { //but also the
resolution and the voxelsize so show an error
                  GateError("GateVImageActor -- Construct: Please give a combination
of two between" <<
                                    " the size, the resolution and the voxelsize (not all) for the sensor");
          }
  }

  if (!mResolutionIsSet && !mVoxelSizeIsSet) { // the size is defined
but neither the resolution nor the voxelsize, so use a default value
    mResolution = G4ThreeVector(1.0, 1.0, 1.0);
    mResolutionIsSet = true;
  }

So now any combination of two of this variables lead to a valid Image Actor.

dsarrut:
Closed in #29

PR #29, opened 2015-06-04
pull request to solve issue #27

BishopWolf:
these are the necessary changes to solve issue https://github.com/OpenGATE/Gate/issues/27
It only affects voxelized actors, now you can create an actor by specifying two of the following three lines (you must comment one), so you are not forced to give the size
/gate/actor/doseDistribution/setSize	Sx	Sy	Sz	mm
/gate/actor/doseDistribution/setResolution	Nx	Ny	Nz
/gate/actor/doseDistribution/voxelsize 	VSx	VSy	VSz	mm

I have test it and it works like a charm for me.


dsarrut:
thanks !

PR #36, opened 2015-06-09
Adding extra option for dumping density image

BishopWolf:
Also some bug fixes
  1. Air was splitted into several materials with increasing densities, now this is avoided, fixed
  2. Air density was averaged with the following material density, wrong, Air density shall be static value, fixed


dsarrut:
thanks !

Issue #28, opened 2015-06-02
Making all Gate Lists Iterable


BishopWolf:
In Gate there are several classes that contain only one data item which is a vector, moreover there is a whole GateListManager class from whom a lot of other classes inherits. These classes doesn't have any way of fast accessing the elements through std:vector functionalities such as iterators. Implementation of this requires a bunch of changes, class by class, but it is required for code optimization since it will speed up element accessing and list traveling. These changes were provided in pull 21, but since they were combined with other issues lose their perspective, so I can provide them as separate pull request if you agree.

PR #38, opened 2015-06-18
Making all Gate Lists Iterable

BishopWolf:
This is the very first step to solve https://github.com/OpenGATE/Gate/issues/28 , if is accepted then more changes will come. All tests runs fine.

### DIFF
--- a/source/general/include/GateIAEAConfig.h
+++ b/source/general/include/GateIAEAConfig.h
@@ -25,7 +25,10 @@ typedef float  IAEA_Float;
 #endif
 
 typedef short IAEA_I16;
-typedef long  IAEA_I32; // RCN changed to long to allow storage of EGS LATCH, Dec. 2006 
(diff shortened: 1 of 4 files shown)


dsarrut:
Sorry Gill, I close this pull-request because:
- it mixes change about Hounsfield process and iterator
- the change in list iterator does not bring new features nor correct bug, and add too much changes (so error prone).

BishopWolf:
Ok, I recognize this leads me annoyed! If you press one by one the commits you will se that there is no references about Hounsfield process and iterator, however global changes reflect it! Whats going on here? 
Changes in list iterator bring the base for future features I planned to add, so it is required.
Example: I want to optimize some classes by grouping variables into structures and making lists of it, but if I don't have a list handler then I cant do that
Other changes require deeper explanation.

BishopWolf:
I create a new clean pull request with only the changes I want to pull, see https://github.com/OpenGATE/Gate/pull/44

PR #47, opened 2015-11-16
Fix crash on deletion of G4Cache object

JStrydhorst:
If a G4SPSRandomGenerator() object is created but never used, it causes a crash when the object is deleted for reasons having to do with the design of the G4Cache class. This doesn't cause a problem for GATE simulations since the problem occurs after the simulation is finished while the GateVSource objects are being deleted, but it does cause the output to end with an error message rather than exiting cleanly.

Moving the creation of mUserPosGenX into InitializeUserFluence() fixes the problem since it only creates the object if it's going to be used.

I haven't tested this with a macro that uses the custom user fluence.

### DIFF
--- a/source/digits_hits/src/GateDigitizer.cc
+++ b/source/digits_hits/src/GateDigitizer.cc
@@ -383,14 +383,14 @@ void GateDigitizer::Digitize()
   // Have the pulses processed by the coincidence sorters
   for (i=0; i<m_coincidenceSorterList.size() ; ++i) {
     if (nVerboseLevel>1)
-      G4cout << "[GateDigitizer::Digitize]: launching coincidence sorter '" << m_coincidenceChainList[i]->GetObjectName() << "'\n";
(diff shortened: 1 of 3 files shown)


dsarrut:
Hello Jared, 

thanks very much ! This it not an easy bug to find ...

Could you please test examples/example_UserFluenceSource before I merge ? 

thanks gain
David

JStrydhorst:
OK. I've tested it with the UserFluence example. I'd managed to just push the problem into the UserFluence section of the code so I've made a few more changes to fix the problem. I also deleted some redundant code that was trying to create mUserPosGenY[i] inside a loop after it had already been created before the loop.

The UserFluence example works correctly.

Fundamentally, the problem is still in the design of the G4Cache, not in the GATE code, so the fix is a kludge. It does nothing but ensure all the G4Cache<> objects inside G4SPSRandomGenerator are created so it doesn't generate an error when it tries to delete the objects.

dsarrut:
thanks !

PR #48, opened 2015-11-26
Timeslice fix

JStrydhorst:
I was having problems when I ran simulations that were too long, i.e. when more than INT_MAX events were required, the later events would have a different runID. If more than 2*INT_MAX events were required, GATE would crash. The problem was caused by the complicated way the time slice information was stored, with the same information stored in different ways in different variables. (a start time, and end time, a vector of end times, and a vector of slice durations.) It made the code bug-prone and hard to read. In fixing the code, I ended up significantly rewriting it to store all the time slice data in a single vector. In all the tests I've run, it seems to work much better. I can run very long simulations with no crashing or wrong runIDs.

I took out the StopDAQ and PauseDAQ functionality because they complicated the StartDAQ function and I couldn't find any way it would ever be possible to call the functions while GATE is running in a single threaded environment. Rewriting the code for multithreading would require some major modification. But maybe I missed something, are they necessary?

### DIFF
--- a/source/general/include/GateApplicationMgr.hh
+++ b/source/general/include/GateApplicationMgr.hh
@@ -57,7 +57,7 @@ public:
   void PrintStatus();
   void SetVerboseLevel(G4int value) { nVerboseLevel = value; }
 
-  G4double GetTimeInterval(int i){return listOfTimeSlice[i];}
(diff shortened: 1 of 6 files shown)


dsarrut:
Thanks a lot Jared. It seems ok from my side, but other should also test (in particular PET and SPECT benchmark)

PR #50, opened 2015-12-14
Dose calculation in inhomogeneous volumes and voxelized phantoms

TomThePhysicist:
There is the DoseActor modifications for inhomogeneous volumes and voxelized phantoms.
With that I have created a new actor who can generate mass images (useful for dose calculation).

The new actor and commands are described in this [wiki](https://github.com/TomThePhysicist/Gate/wiki/Added-things).


BishopWolf:
Dear Tom, This was already done in https://github.com/OpenGATE/Gate/pull/36 but as density image, the mass image is just the same density multiplied by the voxel volume

TomThePhysicist:
Hi Alex, this is an elegant way to generate density image.

If I understand correctly what you have done (I'm not an expert), _buildAndDumpDensityImage_ will generate an image of density corresponding to the resolution of a patient phantom, assuming each voxel contain one material.

But I wonder what is the resulting density image when each voxel contains two or more materials ? If the actor has a bigger resolution than the phantom's resolution for example.

This is what I have tried to resolve with the _VoxelizedMassActor_ in this pull request.

BishopWolf:
I get your point, I just dont want duplication of efforts and code, if there is a consense your way is better so be it, I just think adding an extra resolution option to my way would be enough for your purposes.
On the other hand I am not good with a Mass actor, neither density actor, since this is not an output physical unit, but an input one. so a simple call shall be enough, your way needs to calculate the mass of each voxel, which will be just be done once! an actor shall be used for active things, that evolve with time or simulation, not on fixed quantities. Just my point of view.

TomThePhysicist:
Thank you for your advices !
I will see what I can do about it tomorrow.

PR #51, opened 2016-01-20
Unvoxelized example for dose calculation in inhomogeneous volumes

TomThePhysicist:


TomThePhysicist:
Hello everybody,

I have replaced the initial unvoxelized-example with the gammex by the example with which I have tested the unvoxelized dose algorithm.

It's a simpler example but the user can verify easily if the unvoxelized dose algorithm is working correctly.

dsarrut:
merged, thanks !

PR #55, opened 2016-02-05
Fixes for geant4 10 02

djboersma:
In preparation to the 7.2 release I tried to fix Gate such that it compiles without warnings/errors with clang on MacOSX El Capitan with a NOMT installation of geant4 10.2.

Note that this does *not* yet include my patch to `CMakeLists.txt` to check/warn against MT builds of geant4. That is a separate issue.

I am surprised that apparently stuff got built without problems on other platforms, because there is an include guard bug in geant 4.10.02 which I think should cause all compilers to choke. I inserted an ugly but functional work around, which should be removed as soon as a patch release of geant4 becomes available in which this issue is fixed.

For some weird reason GitHub lists for this pull request also various commits from the past that have already long ago been merged into the official develop branch. I am still not really friends with git.... Only the commits on February 05 are relevant.


djboersma:
Hmm... I notice now that one of my changes is in the same file that pull request #54  (GateVoxelizedMass.cc) is working on. I checked the diffs of that pull request, as far as I can see the one line that I changed does not interfere with those changes. Hopefully git deals intelligently with this.

mojca:
The following seems a tiny bit weird to me:
~~~~~cmake
IF(NOT ${G4VERSION_MAJOR}.${G4VERSION_MINOR} EQUAL 10.2)
    MESSAGE("Warning! GATE version 7.1 is not validated for Geant4 ${G4VERSION_MAJOR}.${G4VERSION_MINOR} distribution. Please use Geant4 10.1 distribution instead.")
ENDIF()
~~~~~

The numbers mismatch, but it would be a bit unfair to issue warnings when using Geant4.10.1 as that one is most likely not deprecated yet. Please note that 4.10.2 requires C++11 which might be a bit of an issue on older OSes. It would be nice if compatibility with Geant4.10.1 was kept a bit longer.

djboersma:
dsarrut wrote this morning that "GATE must use GEANT4 10.2", maybe I overinterpreted that. We could change it such that both 10.1 and 10.2 are allowed.

djboersma:
I reported the geant4 bug (include guards in the Livermore PholoElectric stuff) to the geant4 developers, and Vladimir Ivantchenko kindly told me that *"it turned out that the fix is already done before. The fixed version will be available with the next public patch this month."*
If we wait till this geant4 patch release comes out, then we do not need the ugly workaround. (But we may have to add a check in CMakeLists.txt to encourage the users to upgrade to that patch version, if they haven't already.)

PR #62, opened 2016-03-17
minor modifications to GateRandomEngine

JStrydhorst:
The /dev/random device blocks if the internal estimate of entropy is too low. In practice, this means that when running multiple simulations in succession, as when debugging a macro in interactive mode, there's often a long pause (15-20 seconds on my system) in the random engine initialization step. Using /dev/urandom fixes this trivial but irritating behaviour.

Also modified the code to read a long instead of an int from the supplied seed. Practically, this almost certainly makes no difference whatsoever, but it is in principle more correct, since the variable being set is a long. (For a given macro that supplies its own random seed, it is now possible to run 2^64 different noise realizations instead if 2^32.)

### DIFF
--- a/source/general/src/GateRandomEngine.cc
+++ b/source/general/src/GateRandomEngine.cc
@@ -115,23 +115,23 @@ void GateRandomEngine::ShowStatus() {
 void GateRandomEngine::Initialize() {
   bool isSeed = false;
   long seed = 0;
-  // rest bits are additionnal bit used for engine initialization
+  // rest bits are additional bit used for engine initialization
   // default engine doesn't use it
   int rest = 0;
 
   if (theSeed=="default" && theSeedFile==" ") {
     isSeed=false;
   } else if (theSeed=="auto") {
-    // initialize seed by reading from kernel random generator /dev/random
-    // FIXME may not be protable
-    FILE *hrandom = fopen("/dev/random","rb");
+    // initialize seed by reading from kernel random generator /dev/urandom
+    // FIXME may not be portable
+    FILE *hrandom = fopen("/dev/urandom","rb");
     if(fread(static_cast<void*>(&seed),sizeof(seed),1,hrandom) == 0 ){G4cerr<< "Problem reading data!!!\n";}
     if(fread(static_cast<void*>(&rest),sizeof(rest),1,hrandom) == 0 ){G4cerr<< "Problem reading data!!!\n";}
     fclose(hrandom);
 
     isSeed=true;
   } else {
-    seed = atoi(theSeed.c_str());
+    seed = atol(theSeed.c_str());
     rest = 0;
 
     isSeed=true;


djboersma:
I was going to submit something like this too. I totally agree. And a few months ago I stumbled on this blog post comparing `/dev/random` and `/dev/urandom`, which I found quite instructive:

http://www.2uo.de/myths-about-urandom/

For MC the requirements for random numbers are a bit more relaxed than for cryptography, so I think `/dev/urandom` is completely fine. And much more convenient, because it doesn't block.

djboersma:
However, note that we need to completely rethink this part of the code when we want to enable multithreading in GATE. @BishopWolf is working on this. Maybe I am stating the obvious here.

BishopWolf:
I am sorry, but i cant do this myself atm:

The file is not thread safe as it uses a temporary external file to record the /dev/urandom output. But the change is very simple

include the line
#include <random>

replace the lines
-    // initialize seed by reading from kernel random generator /dev/random
-    // FIXME may not be protable
-    FILE *hrandom = fopen("/dev/random","rb");
-    // initialize seed by reading from kernel random generator /dev/urandom
-    // FIXME may not be portable
-    FILE *hrandom = fopen("/dev/urandom","rb");
-     if(fread(static_cast<void*>(&seed),sizeof(seed),1,hrandom) == 0 ){G4cerr<< "Problem reading data!!!\n";}
-     if(fread(static_cast<void*>(&rest),sizeof(rest),1,hrandom) == 0 ){G4cerr<< "Problem reading data!!!\n";}
-     fclose(hrandom);
 
 -    isSeed=true;
-   } else {
-    seed = atoi(theSeed.c_str());
-    seed = atol(theSeed.c_str());

by the lines
 std::random_device rd; //this uses /dev/urandom by default
 seed = rd(); // Generates a single random int
 isSeed=true;

the result is thread safe and works on all platforms

sorry not to be able to do this for now, but i wonder why this approach were not used from the begining

djboersma:
It depends on what you mean by "from the beginning". The `std::random_device` is only available since C++11, see: http://en.cppreference.com/w/cpp/numeric/random/random_device

BishopWolf:
Oh, then my solution only works for c++11 or up! So I must call for a consensus if we want to do this. I have the pull request ready.
Sorry for my c++11 mind!!

djboersma:
Well, we live in 2016, so maybe it's about time to accept C++11 as the norm, and C++14 as the wild new thing. :)

BishopWolf:
The changes needed are vast:
You have to compile Geant4 with c++11 norm. No big deal ;) (already possible)
You have to compile Root with c++11 norm. No big deal ;)  (already possible)
You have to compile Gate with c++11 norm. No big deal ;) (already possible)

So, what i see is that we are ready for that step, users will complain but i think is for the best since c++11 is the standard norm in most platforms. If we agree i push the request.

djboersma:
I think that the question of which C++ standard to use is an excellent topic to discuss in Orsay (I will attend only from remote at the SC meeting). I think it would indeed be reasonable to accept and use the "new" (five year old) standard, and stop limiting ourselves to C++03. In an earlier discussion I was told that there are Gate users on older linux boxen (like ubuntu 10.04 and 12.04, I guess) who would be unhappy with such a policy. But I think that we do not have resources in this collaboration to keep supporting the very old stuff, and we want to move on, so I'd recommend those users to upgrade to ubuntu 16.04 LTS when it gets released, and they might keep using that until 2020. :)

mojca:
C++11 is a bit problematic on OS X 10.8 and earlier, but Geant 4.10.2 (including the latest version of CLHEP) and ROOT 6 all require C++11 already. So once Gate drops support for Geant 4.10.1, it no longer makes sense to insist in pre-C++11 code. It would however make a lot of sense to make releases of Gate a lot more often, so that users could fetch the latest release that's still supported on their system. At the moment there is basically no clear-cut.

PR #64, opened 2016-03-23
Fix rangen with c++11 standard

BishopWolf:
This was discussed in https://github.com/OpenGATE/Gate/pull/62 so refer there for history.

this allows the random engine to have an automatic seed with thread safe manner.

Main issue here is that it must be compiled with c++11 support


dsarrut:
Hello, 

thanks to report ! However, we cannot (yet) oblige the user to compile with C++11. Could you please add a test that check if it is compiled with c++11 or not, and use the old or your new method accordingly ? 

thanks 
David

BishopWolf:
Added, sorry for the delay, I was in a congress. Can you check that it is compilable with previous c++11 standards?

dsarrut:
sounds good, thanks Alex !

BishopWolf:
Hi https://github.com/matthiasrichter, please refer to https://github.com/OpenGATE/Gate/issues/90 for the c++11 discussion and to the https://github.com/OpenGATE/Gate/pull/44 to see what do we gain by making the jump. Sorry for the delayed answer i wasnt connected at all ;)

PR #65, opened 2016-03-31
Headerfile sanitation

djboersma:
This is a code gardening patch.

When I started looking into slurm support for gjs, I noticed that there were a few header files with `using namespace std` in them. My c++ gurus always told me that `using` directives should never be used in header files. So I tried to clean that up. While doing that I also noticed that some of those header files had `#include`s in them which were not at all needed for the declarations (and/or inline implementations) of the classes/functions in those header files. So I removed those `#include`s and added them wherever they were needed in implementation files. Some namespace prefix issues also had to be fixed. In header files I think we should always use explicit prefixes (we cannot use `using`). In an implementation file I'll only use a `using` directive if the used thing occurs many times in that file.

While doing this I noticed that there is a bit of a mess concerning printing messages to standard output. Gate has its own `GateMessage` infrastructure. But in many places also `G4out`, `std::cout` and/or `std::endl` are used, all of which should be replaced with `GateMessage`, I guess. That should then be done in a separate pull request.

### DIFF
--- a/cluster_tools/filemerger/include/GateMergeManager.hh
+++ b/cluster_tools/filemerger/include/GateMergeManager.hh
@@ -18,13 +18,11 @@ See GATE/LICENSE.txt for further details
 #include <TChain.h>
 #include <cstdlib>
 
-using namespace std;
-
 class GateMergeManager
 {
 public:
 
-  GateMergeManager(bool fastMerge,int verboseLevel,bool forced,Long64_t maxRoot,string outDir){
+  GateMergeManager(bool fastMerge,int verboseLevel,bool forced,Long64_t maxRoot,std::string outDir){
      m_verboseLevel = verboseLevel;
      m_forced       =       forced;
      m_maxRoot      =      maxRoot;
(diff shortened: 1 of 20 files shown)


djboersma:
fwiw: i tested this with gcc 5.3.0 (custom install on scientific linux 6.7) and with gcc 4.8.4 (default compiler on ubuntu 14.04).

BishopWolf:
For you to know, there is already a Gateendl variable that already handles with conventions, so I once sanitize the code by putting all std::endl, endl, and others to Gateendl, but since the code still grows and people are still contributing (so nice), it seems that was lost at some point, can you check that using Gateendl everywhere works for you.

djboersma:
Yes, I noticed the `Gateendl` macro. It is used both within `GateMessage` and with `std::cout`/`G4out`. For the latter does not make much sense to use `Gateend`, in my opinion. The bigger question is whether there is any legitimate reason to prefer `std::cout`/`G4out` over GateMessage anywhere in Gate. I cannot think of any, so I'm inclined to replace them all with `GateMessage` (and then also use `Gateend`, obviously). But as I already wrote, that should be dealt with in a separate pull request.

BishopWolf:
AFAIK, there is a preference to use GateMessage over any other output system, but that is a titanic task. Please make use of this https://github.com/OpenGATE/Gate/pull/17/commits/2b8f5942396baf97c7dd7747cc8be597558240a5

djboersma:
I guess you mean this:

> Convention
>
>  avoid using "string"<<Gateendl, instead use "string\n"
>  avoid using variable<<"\n", instead use variable<<Gateendl
>  avoid using blank spaces before the end of line, they are invisible but increases the sizes of log files
>  avoid using long spaces separators like "    ", instead use tabs like "\t"

I agree with the 2nd, and with the 3rd (for different reasons), I do not understand the 1st and 4th.
1st: Why would `"string\n"` be better than `"string" << Gateendl`?
4th: I don't like tabs so much, because (1) their length is not clearly defined, and (2) whenever the log file gets minor edits, like prepend every line with some characters (quoting stuff in an e-mail) then the tabs lose their alignment, etc. Disk space is cheap, logfiles can be compressed, so spaces are not a concern for log file sizes I think.

Titanic task: a recursive grep in the source tree gives about 2200 hits (checking only .cc and .hh files). That's a lot, but I think that a lot may be done with find & replace. Classifying messages with appropriate logging level could be much more effective to reduce log file sizes than tweaking the amount of whitespace in them.

BishopWolf:
That is a convention, not a rule or an order, there is no gaining between <"string\n"> and <"string" << Gateendl> it is just for code readability, it is better to read the first, but you can use either. What i seek with the convention is that the code shall be consistent, 
Regarding log file sizes, i can show you one of my logs with 500 GB (on my server, hehe), so every effort in avoiding useless things shall be done, tabs on the other hand are just a character that is defined in every system and it is intended for tabular data only, so the convention might be upgraded to dont use spaces to separate tabular data, instead use tabs. and dont use tabs in any other situation.

PR #66, opened 2016-04-14
Coincidence sorter

JStrydhorst:
I think the coincidence sorter can be integrated into the main code. Multiple window mode (allPulseOpenCoincGate) now works and has been validated in takeAllGoods mode.

Some of the multiples policy options haven't been fully validated. 'killAll' and 'keepAll' are trivial and should work fine, 'takeAllGoods' has been tested, the rest should be validated before use. (If they are ever used. Is there ever a use for something like 'keepIfOnlyOneIsGood', which doesn't create a coincidence but still passes the pulse to affect any buffer or dead time modules that follow?)


dsarrut:
Hello Jared ! 

Should I merge now ? no more hardcoded variables ;) ?

thanks, 
David

JStrydhorst:
Yes I think you can do the merge. That comment refers the Y90 code, which I separated from the coincidence sorter branch a while ago. This pull request only affects the coincidence sorter.

PR #67, opened 2016-04-19
Change default random seed in GateToImageCT.cc

kjoshi:
The call to `CLHEP::HepRandom::setTheSeed()` in `GateToImageCT.cc` appears to override the random engine seed to using `/gate/random/setEngineSeed`.
Since `m_seed` in `GateToImageCT.cc` is set to `567665` by default, this can lead to different simulations, with apparently different random engine seeds, producing identical outputs.

The proposed change alters the default behaviour. 
`m_seed` in `GateToImageCT.cc` is now set to `-1`, and `CLHEP::HepRandom::setTheSeed()` is only called if `/gate/output/imageCT/setStartSeed` is explicitly called by the user, with a value different to the default.


dsarrut:
Good catch Kiran !

PR #68, opened 2016-04-20
removed material handling from ParameterizedCollimator

JStrydhorst:
The GateParameterizedCollimator class had a bug that caused the material of the collimator to be set to Vacuum any time one of the parameters (focal length, hold size, dimensions, etc.) was modified. The work-around was to set the material after all of the other parameters were set, but I've now fixed the code so that's not necessary.

The problem was that the GateParameterizedCollimator class had it's own mMaterialName parameter which defaulted do Vacuum and couldn't be modified. Every time one of the other parameters was updated, the ResizeCollimator function used mMaterialName to update the mMaterialName in the parent-class (grand-parent class ?) GateVVolume. Everything in the GateParameterizedCollimator class that refers to mMaterialName has been deleted. There's no need to duplicate functionality already inherited, much less break the functionality of the parent class.

[There's still a lot of weirdness in this code. I don't understand why the 'this' pointer is accessed by GetCollimatorCreator() (which then calls GetCreater()), rather than just accessing the methods of the parent classes directly. The way this code is written, there's nothing to prevent the user from using the commands inherited from BBox to modify the collimator dimensions, though doing that wouldn't update the holes pattern dimensions to match.]

### DIFF
--- a/source/geometry/include/GateParameterisedCollimator.hh
+++ b/source/geometry/include/GateParameterisedCollimator.hh
@@ -55,8 +55,6 @@ class GateParameterisedCollimator : public GateBox
 	{ return m_DimensionX; }
      inline G4double GetCollimatorDimensionY() const
 	{ return m_DimensionY; }
-     inline const G4String& GetCollimatorMaterial() const
-	{ return mMaterialName; }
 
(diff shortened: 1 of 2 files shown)


dsarrut:
Thanks Jared !

PR #71, opened 2016-05-25
Add fastY90 model to GATE

JStrydhorst:
This code implements fast simulation of an Y90 source, creating the bremsstrahlung photons directly without simulating the electron transport of the beta emission. The method implemented is that of Rault et al (Med Phys 37:2943). Voxelized sources are supported for the Interfile format only.


dsarrut:
Thanks Jared !

PR #78, opened 2016-07-01
Abort on critical errors.

patayg:
Current code implementation is full of "safe" guesswork which turns debugging unnecessarily difficult.
If an include file is not present, only a log message is emitted:
```ERROR: Can not open a macro file <name_of_the_macro>. Set macro path with "/control/macroPath" if needed.```

If there is an unresolved alias, e.g.
```Alias <zposition> not found -- command ignored```
```***** Illegal parameter (0) </gate/source/sourceName/gps/centre 12.0 34.0 {zposition} mm> *****```
```***** Batch is interrupted!! *****```

but the program still continues to run, with a who-knows-what kind of default setup.
In this particular case the default value is zero, which might or might not be a good guess, but it clearly contradicts the intended use case (i.e. to control a parameter of the simulation).

Such silent errors shall terminate the simulation immediately, with an error message and a nonzero return value.

### DIFF
--- a/Gate.cc
+++ b/Gate.cc
@@ -20,6 +20,7 @@
 
 #include "G4UImanager.hh"
 #include "G4UIterminal.hh"
+#include "GateUIterminal.hh"
 #include "G4UItcsh.hh"
 #include "GateRunManager.hh"
 #include "GateMessageManager.hh"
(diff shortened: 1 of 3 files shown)


dsarrut:
totally agree with that, thanks !

PR #79, opened 2016-07-11
Fixed relative path to visu.mac in example_fastY90

canonnervio:
The relative path to visu.mac was missing the 'mac/' folder prefix


dsarrut:
thanks !

Issue #74, opened 2016-06-06
dose actor attached to patient CT data (ImageNestedParametrisedVolume)


djboersma:
When I tried to configure my dose actor such that the dosel geometry would be the same as the patient voxel geometry by setting the same voxelsize, I got a 511x511x55 dose distribution while the patient is actually 512x512x55, and the "elementspacing" in the dose MHD file does not match the voxelsize which I configured, and it is also different from the patient voxelsize; I think that the only quantity that was preserved is the total size of the image volume (I think you call that the "bounding box"). In Z the spacing is a round number, namely 3mm, but in X and Y it's 1.03516mm, so I'm guessing there is some rounding issue going on. If instead of the VoxelSize I set the Resolution to 512x512x55 this problem seems to be solved (or rather: worked around).

I have browsed a bit through the dose and image actor code but it's quite a deep rabbit hole, so I can't really claim yet that I found a bug. I can look into it later (and learn a lot about ITK and imaging). But maybe also someone else can have a look at this who has more experience with imaging code and has a clearer idea where/why this might happen. I can share the details for reproducing the problem, if needed.

Another solution which I tried is to omit both voxel size and resolution. I was hoping that it would take those settings from the patient data, but instead the dose actor regarded the entire patient image as a single voxel. I thought that was odd.

I was going to clarify the DoseActor documentation on the wiki, based on these experiences, but after some more pondering I thought that this is actually something that needs fixing, or at least discussion.

So this issue has two parts:

(1) There seems to be a rounding problem with `setVoxelSize` in the dose actor.
(2) Why doesn't the dose actor by default use the same voxelization as the patient data?

TomThePhysicist:
Hello,
This is a known issue which, I think, does not depend on dose actor.
Can you try the branch dicom-dev and tell me if the problem remain ?
Thomas

djboersma:
Hi Tom, thanks for pointing me to `dicom-dev`, I'll try it out! Probably also very useful for many other things...

BishopWolf:
Refer to https://github.com/OpenGATE/Gate/pull/29 for discussion. The line that does this is the 155 in the GateVImageActor.cc. So you want to change the default behaviour from

mResolution = G4ThreeVector(1.0, 1.0, 1.0);

But this requires a definition tweak, the Voxel actor can be used in ANY kind of geometry, not only voxelized ones, so what is "the same voxelization as the patient data" in a case where the are no patient at all? I think current behaviour is more general

[...]

PR #82, opened 2016-08-12
Adding DICOM images support to GATE

TomThePhysicist:
But also correct GATE rounding issues with MHD images (issue #74).

For now it **need an external ITK** (>4.9) to work and **only support the reading** of DCM images.


TomThePhysicist:
Hello,

For more informations you can look at the [link](http://hypernews.slac.stanford.edu/HyperNews/geant4/get/geometry/1243/1/1/1.html) in the commit description.

I think this bug is linked with the nested parameterization (its voxel resolution and its placement) and the placement of the source.
I have triggered the bug with a nested parameterized volume with a pair resolution of voxels (40 20 80), a placement at 0 0 0 and a isotropic point source placed at 0 0 0.
And I have bypassed this bug by translating the volume with an arbitrary small length.

I don't know if this bug is already reported but it is known since at least 2012 and it is still present in the version 4.10.2.p01.

Since it is related to the nested parameterization, I think it can be tricky to develop a workaround directly in Gate.

I hope I have helped,

Thomas

djboersma:
Ah, oops, somehow I overlooked the commit message that link. Sorry for that.

It sounds like the problem happens if a source is sitting exactly at the boundary between voxels, or at the vertices of the voxels. I wonder if the same problem happens if you place eight cubes such that they all touch each other at the origin, and you put a source at the origin. Geant4 should define sensible and robust behavior for such cases....

It is tempting to dive into this and try to find the root of the problem but I do not have time for that. Maybe all we can do for now is to make a note in our wiki to warn the users that this problem exists and needs this simple workaround. If it is already mentioned in the wiki then maybe I looked in the wrong place, I did not find it.

Issue #73, opened 2016-06-01
TPS pencil beam source


djboersma:
Another summer project: update `GateSourceTPSPencilBeam`.

(1) pencil beam initialization

Right now **all** pencil beams in a given plan are initialized at the start of the simulation as `GateSourcePencilBeam` objects. GSPB is not a simple struct, so for big treatment plans with many thousands of spots this takes a lot of RAM. In my current treatment case it's about 2GiB, Hermann told me that in he also saw Gate using 8 GiB due to `GateSourceTPSPencilBeam`. That is unpractical, for instance for simulating many (versions of) plans on a cluster. It would be better to have just a single GSPB object at any point in time. This would also model the reality much closer, where we treat with just a single pencil beam that has the same properties (it does not jump from proton to carbon, or from one set of source properties to another). 

(2) spot specification text file parsing

This is a different problem related to the same code: the text file parsing code is not very robust. The number of comment lines is hard-coded and if the user accidentally inserts or removes comment lines then the rest of the file yields bogus results or is ignored altogether, WITHOUT reporting any errors. This is in particular a problem for users who generate their spot file from non-DICOM input (because for dicom there is the `clitkDicomRTPlan2Gate` utility in `vv`). Minor mistakes can lead to a waste of a lot of time.

I would like to:
(a) allow arbitrary numbers of comment lines
(b) let each line start with a keyword, which may help to catch mistakes

Wish (a) is easy to implement: just replace all the `for (i=0; i<Nlines; i++) getline(buf);` expressions (where `Nline` is some hardcoded number) with a `GateReadLine` function which reads lines of text and returns the first non-empty non-comment line.

Wish (b) is also easy.

Of course there LOTs of other solutions (xml, dicom) but I like my solution because it is much less invasive while still addressing the problems I raise. The old format can still be supported (backwards compatibility), the new format can for instance be triggered by an obligatory keyword in the first comment line of the file (so that first comment line should of course *not* be skipped). And the `clitkDicomRTPlan2Gate` utility in `vv` could be updated to enable the new format (but as long as we keep the TPS code backwards compatible, that's not urgent).

(3) Convergence/divergence

Minor issue: a pencil beam can be convergent in X but divergent in Y and vice versa. At Skandion we see this at the lowest energy, `Ekin=60MeV`. For a single pencil beam source one can configure this, but in the TPS pencil beam source it's currently restricted to completely divergent or completely convergent. It would be nice to flexibilize this.

[...]

PR #84, opened 2016-08-31
TPS pencilbeam issue #73

djboersma:
This pull request tries to address several of the issues raised in issue #73 about the `GateTPSPencilBeam` actor. This is not yet the final version of the pull request, there are some things that I probably want to change/add based on your feedback.

Changes:

1. It replaces the vector of pencil beam objects, one for each spot in the plan file, with a single pencil beam. Instead of randomly sampling from all spots, the spots are simulated in the same order as configured (the number of protons per spot is still random, of course). Every time we move from one spot to the next, the pencil beam object gets re-initialized with the relevant spot position, direction, energy, and spreads on those.
2. The parsing of the PBS plan description file has been rewritten. The number of comment and/or empty lines between the actual content lines is now free, so adding a comment line does not result in wrong results anymore. Also the type and number of items on each content line is checked and a descriptive error message is emitted when something does not seem right

.

Temporary:

The old code for particle generation is still available in the `OldVertexGeneration` method. The new code is in the `NewVertexGeneration` method. The user can now choose between the two methods with a configurable parameter. This is only for development and debugging, for a release we keep only one of these two implementations. I used this to run example5 of the radiotherapy examples with OldStyle on and off with a million protons and then check that the two methods yield statistically compatible distributions for `X`, `Y`, `Z`, `dX`, `dY`, `dZ` and `Ekine` for every single spot. 

Additionally:

1. I tried to make the text output a bit more consistent: use `GateMessage("Beam',lvl,msg << Gateend)` everywhere and avoid `cout`, `G4cout` and `\n`.
2. A small fix in `GatePhaseSpaceActor`, included in this pull request, was needed to enable spot-wise testing.

Not yet done, to be discussed:

1. In issue #73 I proposed (extra wish "b") to add optional keywords at the start of every line, to increase readability and facilitate debugging. If we want this then we should also add this feature to VV's `clitkDicomRTPlan2Gate`.
2. The new vertex generation method generates the protons sorted according configured spots, but for some use cases, in particular visualization (the `--qt` option) it might be nice to also support random sampling. This can be done of course, but since the pencil beam is going to be reconfigured for almost every proton this will be a tiny bit slower. For visualization with low statics this is not a problem, I think.
3. The new generation method obtains the "total number of primaries to be generated" from the `GateApplicationManager` instance. This is fine if there is only one source of primaries which is supposed to generate all those primaries and then Gate stops, which is typically the case for a non-interactive TPS simulation. But if there are several sources then the TPS pencil beam source will probably be asked for a smaller number of primaries, which is bad, because this will mean that some of the spots at the end of the plan will not be sampled. The actor will also not run correctly if in a GUI run the user requests a number of primaries more than once. If these situations (multiple source actors, or multiple runs in a GUI) are valid use cases then there is still work to do on the new version of this this actor (the old version has no problem with these use cases). It is not hard to make these fixes, but before I do it I want to make sure that I understand the potential use cases well enough.
4. I did not yet implement the option to configure convergence/divergence properties for X and Y separately. I think this should be done in the source properties file, with two extra lines at the end. The convergence might depend on energy. In that case I think a reasonably flexible syntax would be that the user specifies whether the lowest possible energy beam is convergent or divergent, as well as a list of energies this flips between one and the other. E.g.:
```
X divergent 50 250
Y convergent
```
In this example the beam is always convergent in Y but for X it is divergent for energies less than 50 and higher than 250 MeV.

### DIFF
--- a/source/digits_hits/src/GatePhaseSpaceActor.cc
+++ b/source/digits_hits/src/GatePhaseSpaceActor.cc
@@ -108,7 +108,7 @@ void GatePhaseSpaceActor::Construct() {
   EnableBeginOfEventAction(false);
 
   // bEnableEmissionPoint=true;
-  if (bEnablePrimaryEnergy || bEnableEmissionPoint) EnableBeginOfEventAction(true);
+  if (bEnablePrimaryEnergy || bEnableEmissionPoint || bEnableSpotID) EnableBeginOfEventAction(true);
 
   EnablePreUserTrackingAction(true);
   EnableUserSteppingAction(true);
--- a/source/physics/include/GateSourcePencilBeam.hh
+++ b/source/physics/include/GateSourcePencilBeam.hh
@@ -11,7 +11,6 @@
 
 #include "GateConfiguration.h"
 
-#ifdef G4ANALYSIS_USE_ROOT
 #include "G4Event.hh"
 #include "globals.hh"
 #include "G4VPrimaryGenerator.hh"
@@ -34,7 +33,6 @@
 #include "CLHEP/Matrix/Vector.h"
 #include "CLHEP/Matrix/SymMatrix.h"
 #include "GateRandomEngine.hh"
-#include "TMath.h"
 
 class GateSourcePencilBeam : public GateVSource, G4UImessenger
 {
@@ -125,5 +123,4 @@ protected:
   int mCurrentParticleNumber;
 };
 
-#endif
 #endif
--- a/source/physics/include/GateSourcePencilBeam.hh
+++ b/source/physics/include/GateSourcePencilBeam.hh
@@ -76,8 +76,10 @@ public:
   //Correlation Position/Direction
   void SetEllipseXThetaArea(double EllipseXThetaArea) {mEllipseXThetaArea=EllipseXThetaArea;}
   void SetEllipseYPhiArea(double EllipseYPhiArea) {mEllipseYPhiArea=EllipseYPhiArea;}
-  void SetEllipseXThetaRotationNorm(std::string rotation) {mEllipseXThetaRotationNorm=rotation;}
-  void SetEllipseYPhiRotationNorm(std::string rotation) {mEllipseYPhiRotationNorm=rotation;}
+  void SetEllipseXThetaRotationNorm(std::string rotation) {mConvergenceX=(rotation=="positive");}
+  void SetEllipseYPhiRotationNorm(std::string rotation) {mConvergenceY=(rotation=="positive");}
+  void SetConvergenceX(bool b) {mConvergenceX=b;}
+  void SetConvergenceY(bool b) {mConvergenceY=b;}
   void SetTestFlag(bool b) {mTestFlag=b;}
 
 protected:
@@ -108,8 +110,8 @@ protected:
   //Correlation Position/Direction
   double mEllipseXThetaArea;	//mm*rad
   double mEllipseYPhiArea;	//mm*rad
-  std::string mEllipseXThetaRotationNorm;
-  std::string mEllipseYPhiRotationNorm;
+  bool mConvergenceX; // true corresponds to: X-Theta rotation norm is positive
+  bool mConvergenceY; // true corresponds to: Y-Phi rotation norm is positive
   //Gaussian distribution generation for direction
   RandMultiGauss * mGaussian2DYPhi;
   RandMultiGauss * mGaussian2DXTheta;
--- a/source/physics/include/GateSourcePencilBeam.hh
+++ b/source/physics/include/GateSourcePencilBeam.hh
@@ -51,41 +51,41 @@ public:
   void GenerateVertex( G4Event* );
 
   //Particle Type
-  void SetParticleType(G4String ParticleType) {strcpy(mParticleType, ParticleType);}
-  void SetWeight(double w) {mWeight=w; }
+  void SetParticleType(G4String ParticleType) {mIsInitialized &= (ParticleType==mParticleType); mParticleType = ParticleType;}
+  void SetWeight(double w) { mWeight=w; } // no need for re-initialization
   double GetWeight() {return mWeight; }
   //Particle Properties If GenericIon
   void SetIonParameter(G4String ParticleParameters);
   //Energy
-  void SetEnergy(double energy) {mEnergy=energy;}
-  void SetSigmaEnergy(double sigmaE) {mSigmaEnergy=sigmaE;}
+  void SetEnergy(double energy) {mIsInitialized &= (mEnergy==energy); mEnergy=energy;}
+  void SetSigmaEnergy(double sigmaE) {mIsInitialized &= (mSigmaEnergy==sigmaE); mSigmaEnergy=sigmaE;}
   //Position
-  void SetPosition(G4ThreeVector p) {mPosition=p;}
-  void SetSigmaX(double SigmaX) {mSigmaX=SigmaX;}
-  void SetSigmaY(double SigmaY) {mSigmaY=SigmaY;}
+  void SetPosition(G4ThreeVector p) {mPosition=p;} // no need for re-initialization
+  void SetSigmaX(double SigmaX) {mIsInitialized &= (mSigmaX==SigmaX); mSigmaX=SigmaX;}
+  void SetSigmaY(double SigmaY) {mIsInitialized &= (mSigmaY==SigmaY); mSigmaY=SigmaY;}
   //Direction
-  void SetSigmaTheta(double SigmaTheta) {mSigmaTheta=SigmaTheta;}
-  void SetSigmaPhi(double SigmaPhi) {mSigmaPhi=SigmaPhi;}
+  void SetSigmaTheta(double SigmaTheta) {mIsInitialized &= (mSigmaTheta==SigmaTheta); mSigmaTheta=SigmaTheta;}
+  void SetSigmaPhi(double SigmaPhi) {mIsInitialized &= (mSigmaPhi==SigmaPhi); mSigmaPhi=SigmaPhi;}
   // first rotation possibility => Necessary for the GateSourceTPSPencilBeam !!!!!!
-  void SetRotation(G4ThreeVector rot) {mRotation=rot;}
+  void SetRotation(G4ThreeVector rot) {mRotation=rot;} // no need for re-initialization
   //second rotation possibility
-  void SetRotationAxis(G4ThreeVector axis) {mRotationAxis=axis;}
-  void SetRotationAngle(double angle) {mRotationAngle=angle;}
+  void SetRotationAxis(G4ThreeVector axis) {mRotationAxis=axis;} // no need for re-initialization
+  void SetRotationAngle(double angle) {mRotationAngle=angle;} // no need for re-initialization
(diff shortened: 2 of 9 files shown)


djboersma:
Forgot to mention: there were `G4ANALYSIS_USE_ROOT` ifdefs in the pencil beam code, but the code does actually not at all depend on ROOT. So I removed those. Also: I removed the include guards from the implementation files. Include guards really only make sense for header files.

djboersma:
Also: in my own actual simulations with a large number of spots the new generation algorithm (with a single pencil beam instead of a whole vector) seems to reduce the memory footprint by about one GiB, from 4.5 down to 3.5 GiB. I think that there is still a lot to gain in RAM reduction, for instance by using sparse matrices in the dose actor. I guess that should be posted as a separate "issue".

dsarrut:
Hi !

first thanks really a lot for this proposal ! Regarding the proposed changes, it is perfect from my side. The idea to manage only one PB object instead of a list was planned long time ago but never done. The parsing is now ok. I propose to remove the oldVertex method, if you found that example5 lead to the same result.

Points to be discussed:
1. why not but it requires to change clitkDicomRTPlan2Gate. Anyway, this external clitkDicomRTPlan2Gate is not convenient. I will propose to start an action towards providing convenient command line executable within Gate source code. I will explain it as a separate issue. So, in the meantime, I propose to not change.

2. I may disagree here: I think random sampling *must* be the default way. It allows to start simulation (with ou without GUI) and obtain an estimate of the result whatever the number of particles is simulated. Moreover, in some situation (with the [gatelab](http://gatelab.creatis.insa-lyon.fr) ) simulations are sometimes stopped before they reach the initial asked nb of particles (for performance reason, see [doi](http://dx.doi.org/10.1007/s10723-010-9153-0) and [pdf](https://www.creatis.insa-lyon.fr/site7/en/publications/CAMA-10)). However, an option may propose the user to use sorted spots. I understand that it could be a bit slower with random sampling (reconfiguration + random shooting), but do you have an estimate of this overhead ?

3. Exact, multiple sources could be a valid use case. I propose to keep it simple from now, by only checking the nb of source and warn the user if not == 1.

4. this point should be discussed with Loic and Alessio @aelia21 

From now, I merge but leave the thread open for further comments.
thanks again, 
David

dsarrut:
by the way, did you check the line I add recently : 
https://github.com/OpenGATE/Gate/blob/develop/source/physics/src/GateSourceTPSPencilBeam.cc#L527
before your changes, it allowed to save some memory ...

djboersma:
I think it is relatively straightforward to enable random sampling in the new method. There may be some extra overhead though. When simulating in spot-sorted order, the pencil beam only needs to be re-initialized between different layers (because the parameters that require initialization when changed all depend on energy, changing the spot position within a layer does not change the spot profile so it does not trigger a re-initialization). When sampling randomly, successive protons will usually be in spots from different layers and therefore require re-initialization of the pencil beam object. However, I suspect that this "extra overhead" is probably insignificant relative to the overall simulation time.

djboersma:
And your recently added line is also present in my patch. Spots with zero weight should of course be skipped.

djboersma:
In the long term, simulating protons in configured order of the spots (instead of randomly) would enable adding the treatment time in the simulation. For this to happen the spot specification needs to get a time window, and for an actually interesting use of this we would need a 4D patient model, e.g. based on images from 4DCT combined with a breathing curve.

aelia21:
Hello,

I would like to add some comments as well:

1 - from my side, the modification proposed at Point 4 from djboersma ( beam divergent and convergent in two different planes independently) is not needed at the moment. Of course, in term of flexibility of the code it would be a nice improvement anyhow.

2 - I did a very small modification as well in the GateSourceTPSPencilBeam.cc  (available here https://github.com/OpenGATE/Gate/pull/86) and I wanted make you guys aware. Basically the energy spread it should be evaluated in percentage according to what it is already written in the wiki.

3 - I think we could add to the "to-do-list" also the implementation of the generic ion feature which at the moment is working only for carbon ions. I will work on this. 

Last but not least I am really thankful to djboersma for the very good job he did on the code. 

Best
Alessio

djboersma:
Hi Alessio, thanks for reviewing my pull request!
1. It's not a major priority for us either; for practically all energies the beam is convergent in both X and Y, only for the lowest energy beam we saw that it was very slightly divergent. On the other hand most of my simulations use a range shifter, which washes out this subtlety.
2. I commented on your pull request. I think it needs some discussion.
3. Yes, that seems like a very natural extension. When I grep for "proton" in the code I see that in most cases it does not matter whether it's for protons or ions, so a simple name change will do. But the particle type and the `ConvertMuToProtons` function need real work. That function has hardcoded constants in it, so it is anyway a good idea to recode that in a way that allows to make this flexible enough that can be used for any PBS device. At Skandion a Fermi cup calibration was done and the result almost exactly matches the convertMuToProtons function, so we were lucky here.

I suggest that you promote your item 3 (generalize TPSpencilbeam to generic ions) to a new GitHub issue.

brenthuisman:
@djboersma : regarding your benchmark, that memory usage surprises me a little. With a plan of about 500 spots I see memory usage of well below 1GB (I estimate memory allocated for the sources to be about 100-200MB, by taking total Gate mem usage and subtracting geometry which I can compute). Could you comment? (And is there a difference in parsing speed? I recall @dsarrut  and I improved loading speed quite a bit a few years ago, down to a few seconds).

@aelia21 For point 3, are you sure GenericIon (which will actually shoot carbon ions) works at the moment? I used it once, with a proton source definition file and a single spot, but I got a very strange and wide beam (like at least 5cm wide whereas the proton source was indeed just a spot).

brenthuisman:
Because I was curious, I benchmarked NewGenerateVertex() myself:

- new : 3.400M
- real    1m52.542s
- user    1m49.764s
- sys     0m2.764s

- old   : 3.693M
- real    2m0.637s
- user    1m57.088s
- sys     0m3.344s

So, some memory savings and a small speedup. Nice!

djboersma:
I am glad to get some benchmarks!

I should redo my own benchmark a bit more thoroughly. I have to admit that I just looked at `htop` while my test job was running. I think 500 spots is a relatively small plan; the "problem plans" have several thousand spots, my test plan has about 3500 spots, I think that Erik A. and Hermann F. have worked on even bigger plans.

I did not expect any significant gains in simulation speed, apart from speed gains due to the lower memory footprint. I think that CPU time spent on parsing is negligible, both in the old and the new code.

I'm a bit busy with other things this week. Maybe I'll have time for this during the weekend.

brenthuisman:
OK, that explains the difference. My biggest plan has about 1200 spots, this one (field of one plan) about 500. The tiny speedup must be from the initialization. I was worried setting properties of mPencilBeam (for every primary) would be slower than generating mPencilBeams and simply looping over them, but that appears to be no problem.

aelia21:
Hello, sorry for the late reply.

@brenthuisman the GenericIon source is working good as far as I know. Concerning your wide spot I have to say that I am not that surprised as you were using a proton source description.

GrevillotL:
Hi guys,

I just wanted to make sure that we kept also the random sampling method as one of the 2 possibility to simulate the PBS delivery?

djboersma:
Hi @GrevillotL : yes, I think that David S also requested that. I think that is easy to implement also in the new vertex method. Need to find/allocate some time to actually do this.

GrevillotL:
Thanks Dave!
I think it is very important to Keep every functionality in PBS and TPS sources available like it was before (like I developped it during my PhD).
Please only add new Features and do not remove old ones.
This is very improtant for us, because we implemented it in a specific way and for some specific reasons.
Best,
LG

djboersma:
I did not remove the old feature: the old and new implementation of the vertex generation are now side by side in the code (for debugging convenience), so by selecting "old style" you currently get exactly the same behavior as previously.

But there is also a request by DavidS to remove the old style code. So I need to carry over (or mimic) the random sampling feature to the new vertex generation method, to make indeed sure that all the existing functionality remains available, and also that the "default" behavior is compatible with how it behaved the 7.2 release.

Could you detail the specific reasons for random sampling? The only practical reason I could think of is for visualization, where Gate only shows the particles from the last N primaries (e.g. N=100) and you want to get an impression of the whole field. I guess there must be more reasons. I would like to be aware of those other reasons so that I can write a to-the-point documentation string to the option that selects between random and ordered sampling. :)

GrevillotL:
Hi,
Thanks for clarification.
Once you remove the old style code and convert it into the new one, probably we can help with Alessio to validate the behavior is as before in release 7.2.
Random sampling makes sense with respect to Monte Carlo method (sampling based on pdf) and allows to access the full dose distribution at anytime. If one does not know how many particles is needed to have enough statistics, then one can send a huge number of primaries and based on the statistical fluctuations scored at regular time intervals one can stop the 3D calculation, at anytime. Sampling all spots delivered, also do not mean  sampling all Protons delivered, because in reality a spot is made of numerous Protons and a spot is also decribed by pdf in Terms of optics and energy distributions.... so it is all about probabilities.
During  my PhD thesis we tested spot by spot sampling for a fixed number of particles per spot against random sampling. For the spot by spot delivery each spot was weighted by its MU and naturally the random sampling had a faster convergence, which actually makes full sense. To make an efficient spot by spot delivery, the number of protons simulated for each spot shall be proportional to its MU and as typically the number of Protons per spot can vary by about 3 orders of magnitudes in a plan, then of course the low weighted spots will not have many particles simulated. This behavior is actually very similar to the random sampling. In the end, the purpose of a non-random sampling is to Keep time dependence of the beam delivery, which is for you an important Topic in order to correalte it with 4D Images. 
I hope this helps!
I could also help to provide some text in the wiki documentation, I guess it is more important to have it there?

brenthuisman:
The difficulty with NewGenerateVertex() and random sampling is that two PDFs are necesary: one per energy layer (because the PencilBeam is set per energy layer) and then within the layer a PDF for the spots. I thought of helping out a little because of the memory savings, but I didn't see an simple way for such nested PDF's, and on top of that I notice that TPSes cant be trusted to produce perfectly correct MU/MSW sums. With that I mean that when you add up the MUs of all the spots in a layer, you'll get a different number than the plan lists for that layer. Idem when I add layers for the field-total, plan-total. Has any of you seen this too?

djboersma:
Thanks for trying to help out! :)

About the PDFs: I do not see the need for nesting. I was going to sample from the PDF of all spots (from all layers), very similar to how it's done in `OldGenerateVertex`. This will introduce some overhead (re-init for every proton) but I am hoping that this will be negligible.

About perfect MU/MSW sums: I think I checked this a while ago for the plans that I have and it seemed correct. I remember that you have to pay attention to whether it's really the MSW per layer or the cumulative MSW. If I find the time for this I'll check it again.

brenthuisman:
I hope that overhead is acceptable too ;)

Yes, the files stores cumulative MU/MSW each layer, so you must do some substraction. The differences I see are small (up to 0.7MU per plan), but every single plan I have does have those small differences, so, just like how OldGenerateVertex only considers spot-weights, NewGenerateVertex should do the same (and your method would, so we're OK).

I'll ask @dsarrut to push a few unrelated changes to TPSPencilBeam to Github.

PR #86, opened 2016-09-21
in GateSourceTPSPencilBeam.cc energy spread evaluated in percentage

aelia21:
energy spread evaluated now in percentage

### DIFF
--- a/source/physics/src/GateSourceTPSPencilBeam.cc
+++ b/source/physics/src/GateSourceTPSPencilBeam.cc
@@ -286,16 +286,13 @@ void GateSourceTPSPencilBeam::GenerateVertex( G4Event *aEvent ) {
 
               //Particle Type
               Pencil->SetParticleType(mParticleType);
-              //Energy
+              
+	      //Energy
               Pencil->SetEnergy(GetEnergy(energy));
-              Pencil->SetSigmaEnergy(GetSigmaEnergy(energy));
+              Pencil->SetSigmaEnergy(GetSigmaEnergy(energy)*GetEnergy(energy)/100.);
 
-              //cerr << "Brent " << GetSigmaEnergy(energy) << " en " << GetEnergy(energy) <<endl;
-
-              //changed because obiously incorrect.
-              //Pencil->SetSigmaEnergy(GetSigmaEnergy(energy)*GetEnergy(energy)/100.);
+              
               //Weight
-
               if (mSpotIntensityAsNbProtons) {
                 NbProtons = SpotParameters[2];
               } else {


djboersma:
Sorry for the late response!

Currently all our scripts and code (we = Uppsala: Erik A. and me) are assuming that the energy sigma must be provided as an absolute number, not as a percentage. I don't know how many other users there are for this actor, but in general I think it is risky to change the interpretation of such a configuration parameter. It's an API change, but one which lets the users scripts run without crash, while delivering subtly different results. For a 100 MeV beam, the numbers for relative and absolute are the same, but for a 200 MeV they differ by a factor of 2. The effect of that can be significant but not dramatic, so it is very easy to misinterpret the results, which is worrisome.

Your pull request erases some comments made by Brent Huisman, who seemed to strongly prefer the absolute value over the percentage. On the other hand the IBA engineers at Skandion seem to prefer to specify this quantity as a percentage. So I think that there is variation in preferences of users (and developers) and rather than forcing one particular choice I suggest that we allow both choices, by making this choice very explicit: add one boolean option, a flag that specifies whether the energy sigma value should be interpreted as absolute or relative. This can be implemented in various ways: (1) as another `G4UIcmdWithABool` in `GateSourceTPSPencilBeamMessenger`, or (2) as a flag in the `SourceProperties.txt` file. Option (1) is easier to implement (though there will be discussion about what the default value should be...), while (2) is maybe conceptually / design-wise better, because it is the provider of the source properties file who really knows (or should really know) whether the energy sigma polynomial describes the absolute or relative sigma.

aelia21:
Hello! 

Actually I did not want to force a choice. Simply I wanted to be consistent with what Loic originally developed and also reported in the [wiki](http://wiki.opengatecollaboration.org/index.php/Users_Guide_V7.2:Source#TPS_Pencil_Beam_source) (see small section "about the source_description_file").
The change (and the comment which I deleted) was actually from Hermann Fuchs and not from @brenthuisman and it was changed simply for a small misunderstanding.
Anyway, beside this small novel, I think we should simply be aligned on this point. If you think is better to let the user free to choose, I am fine with that.

brenthuisman:
I don't work with actual beam models, just fake ones, so I don't really care either way, but as @djboersma says, current source definitions assume absolute numbers, and we shouldn't break them. I favor @djboersmas option two: specify the type in the sourcefile, and parse accordingly. Then you just have to read to sourcefile to understand what's going on instead of looking through Gate source code, mac files with possible new commands and the sourcefile.

dsarrut:
Hi guys !

I think you all agree to provide 2 options, with flag within the plan file. Default option could be % as this is what was initially proposed. That's fine with me ! The next question is now: who is doing the change ;) ? @aelia21  or @djboersma, other ? Just agree before starting to avoid double work, and pull-request when ready !

thanks !

GrevillotL:
Hello guys,
I am starting my way on GitHub... 
I fully agree with David, better to provide the 2 options, i.e. keep the initial developments in % which are working for us since 2010 and add the possibility in absolute numbers as preferred from Uppsala side.
I would suggest the change to be made from Uppsala side, to facilitate the Validation of the new Features. Would you agree David B. ?
Thx,
Loïc

djboersma:
@GrevillotL : OK, I can do this.

But as @brenthuisman  noted, setting the %-option now as the default is a change with respect to the 7.2 release (and probably also some prior releases), where the energy spread is expected to be given in MeV, so as an absolute standard deviation, not a relative one. Our code in Uppsala assumes the spread is specified in MeV. On the other hand there are users (like you and @aelia21 ) who are actually expecting the old behavior, so keeping the current default is just as confusing as changing it. There is no clean solution. Whatever we choose (keep absolute, or revert to %), we need to make a loud and clear note about this in the release notes of the next release.

GrevillotL:
Hello again,
that's a big day for my start on github! I am not sure I can continue with this rythm next week :)....

Basically, I think the old behavior shall not have been modified from the beginning, because it was working and described as such in the documentation - of course everybody is welcome to put new Feature, but in General I think we should be more careful before applying changes on working things. Let's imagine that next year a new PhD Student starts again to Change everything again because I am sure for his application he would have good reasons as well...
In the end both features are correct and I do not care so much about the Default one, but I care just to have both available, so that we can Keep working with our tools.
Thanks,
LG

brenthuisman:
A better specified format could allow users to add their fields without breaking things for others. It seems to me that all Actor flags should ideally be described by the source properties file (particle species, divergence, etc), not macro commands.

Fortunately the ad-hoc format is dead simple, each value is on its own line (which are strictly ordered). Proposal: @aelia21 rewrites the pull request to have units at the end of each value line. No unit means: parse as we used to. Simple fix that doesn't break any current source properties files.

dsarrut:
See #117

PR #95, opened 2016-10-19
Include Compute Capabilities for newer GPUs

ekleung:
Updated CMakeLists.txt to include Compute Capabilities for all NVIDIA CUDA GPUs (up to GeForce GTX 10-series GPUs with Compute Capability 6.1)

Otherwise, Gate/examples/gpumacros/pet/main-gpu.mac would crash when executed.

Also added
Gate/examples/gpumacros/pet/output-cpu/.gitkeep
and
Gate/examples/gpumacros/pet/output-gpu/.gitkeep
Otherwise, the output files cannot be generated.


dsarrut:
thanks

PR #101, opened 2017-02-17
 Correction of the 'HU indices out of range' bug (with ImageRegionalizedVolume)

TomThePhysicist:
The bug was triggered when GateVImageVolume::LoadImage(bool add1VoxelMargin) was used with add1VoxelMargin = true.


dsarrut:
Thanks Thomas !

PR #106, opened 2017-03-15
Modify GateDoseActor.cc

uhqd:
Conversion in gray is made after DebugMessage
Photon is added to the list of particles in the computation of DOSE TO WATER

### DIFF
--- a/source/digits_hits/src/GateDoseActor.cc
+++ b/source/digits_hits/src/GateDoseActor.cc
@@ -409,7 +409,7 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
       }
 
       GateDebugMessage("Actor", 2,  "GateDoseActor -- UserSteppingActionInVoxel:\tdose to water = "
-                       << G4BestUnit(doseToWater*gray, "Dose to water")
+                       << G4BestUnit(doseToWater*gray, "Dose")
                        << " rho = "
                        << G4BestUnit(density, "Volumic Mass")<< Gateendl );
     
--- a/source/digits_hits/src/GateDoseActor.cc
+++ b/source/digits_hits/src/GateDoseActor.cc
@@ -360,13 +360,16 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
   if (mIsDoseImageEnabled) {
     // ------------------------------------
     // Convert deposited energy into Gray
-    dose = edep/density/mDoseImage.GetVoxelVolume()/gray;
+    dose = edep/density/mDoseImage.GetVoxelVolume();
     // ------------------------------------
 
     GateDebugMessage("Actor", 2,  "GateDoseActor -- UserSteppingActionInVoxel:\tdose = "
 		     << G4BestUnit(dose, "Dose")
 		     << " rho = "
 		     << G4BestUnit(density, "Volumic Mass")<< Gateendl );
+	
+	// Convert to Gy AFTER debug message
+	dose = dose / gray;
   }
 
   double doseToWater = 0;
@@ -389,18 +392,19 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
       double Volume = mDoseToWaterImage.GetVoxelVolume();
 
       // Other particles should be taken into account (Helium etc), but bug ? FIXME
-      if (PartName== "proton" || PartName== "e-" || PartName== "e+" || PartName== "deuteron") {
+      if (PartName== "proton" || PartName== "e-" || PartName== "e+" || PartName== "deuteron"|| PartName== "gamma") {
         //if (PartName != "O16[0.0]" && PartName != "alpha" && PartName != "Be7[0.0]" && PartName != "C12[0.0]"){
 
         DEDX = emcalc->ComputeTotalDEDX(Energy, PartName, material, cut);
         DEDX_Water = emcalc->ComputeTotalDEDX(Energy, PartName, "G4_WATER", cut);
-        doseToWater=edep/density/Volume/gray*(DEDX_Water/1.)/(DEDX/(density*e_SI));
+        doseToWater=edep/density/Volume*(DEDX_Water/1.)/(DEDX/(density*e_SI));
         if (DEDX_Water == 0 || DEDX == 0) doseToWater = 0.0; // to avoid inf or NaN
       }
-      else {
+      else { 
+      // FIX ME FOR OTHER PARTICLES
         DEDX = emcalc->ComputeTotalDEDX(100, "proton", material, cut);
         DEDX_Water = emcalc->ComputeTotalDEDX(100, "proton", "G4_WATER", cut);
-        doseToWater=edep/density/Volume/gray*(DEDX_Water/1.)/(DEDX/(density*e_SI));
+        doseToWater=edep/density/Volume*(DEDX_Water/1.)/(DEDX/(density*e_SI));
         if (DEDX_Water == 0 || DEDX == 0) doseToWater = 0.0; // to avoid inf or NaN
       }
 
@@ -408,6 +412,10 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
                        << G4BestUnit(doseToWater, "Dose to water")
                        << " rho = "
                        << G4BestUnit(density, "Volumic Mass")<< Gateendl );
+    
+   	// Convert to Gy AFTER debug message
+	dose = dose / gray;
+
     }
 
   if (mIsEdepImageEnabled) {
--- a/source/digits_hits/src/GateDoseActor.cc
+++ b/source/digits_hits/src/GateDoseActor.cc
@@ -360,16 +360,15 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
   if (mIsDoseImageEnabled) {
     // ------------------------------------
     // Convert deposited energy into Gray
-    dose = edep/density/mDoseImage.GetVoxelVolume();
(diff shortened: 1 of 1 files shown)


source/digits_hits/src/GateDoseActor.cc
```diff
@@ -389,25 +392,30 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
...
         if (DEDX_Water == 0 || DEDX == 0) doseToWater = 0.0; // to avoid inf or NaN
       }
 
       GateDebugMessage("Actor", 2,  "GateDoseActor -- UserSteppingActionInVoxel:\tdose to water = "
                        << G4BestUnit(doseToWater, "Dose to water")
                        << " rho = "
                        << G4BestUnit(density, "Volumic Mass")<< Gateendl );
+    
+   	// Convert to Gy AFTER debug message
+	dose = dose / gray;
```
TomThePhysicist: I think you wanted to write doseToWater instead of dose.

source/digits_hits/src/GateDoseActor.cc
```diff
@@ -360,13 +360,16 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
   if (mIsDoseImageEnabled) {
     // ------------------------------------
     // Convert deposited energy into Gray
-    dose = edep/density/mDoseImage.GetVoxelVolume()/gray;
+    dose = edep/density/mDoseImage.GetVoxelVolume();
     // ------------------------------------
 
     GateDebugMessage("Actor", 2,  "GateDoseActor -- UserSteppingActionInVoxel:\tdose = "
 		     << G4BestUnit(dose, "Dose")
```
TomThePhysicist: You can simply change `G4BestUnit(dose, "Dose")` to `G4BestUnit(dose*gray, "Dose")` here.

source/digits_hits/src/GateDoseActor.cc
```diff
@@ -389,25 +392,30 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
...
+      // FIX ME FOR OTHER PARTICLES
         DEDX = emcalc->ComputeTotalDEDX(100, "proton", material, cut);
         DEDX_Water = emcalc->ComputeTotalDEDX(100, "proton", "G4_WATER", cut);
-        doseToWater=edep/density/Volume/gray*(DEDX_Water/1.)/(DEDX/(density*e_SI));
+        doseToWater=edep/density/Volume*(DEDX_Water/1.)/(DEDX/(density*e_SI));
         if (DEDX_Water == 0 || DEDX == 0) doseToWater = 0.0; // to avoid inf or NaN
       }
 
       GateDebugMessage("Actor", 2,  "GateDoseActor -- UserSteppingActionInVoxel:\tdose to water = "
                        << G4BestUnit(doseToWater, "Dose to water")
```
TomThePhysicist: Same here, you can simply change `G4BestUnit(doseToWater, "Dose")` to `G4BestUnit(doseToWater*gray, "Dose")`.

source/digits_hits/src/GateDoseActor.cc
uhqd: Ok. It is the same. But previous version was wrong because conversion was made twice.

source/digits_hits/src/GateDoseActor.cc
uhqd: ok too

source/digits_hits/src/GateDoseActor.cc
uhqd: yes sorry for this. doseToWater !  3 Lines, 1 error... It is totally me... But most important. Do you agree with the adding of photon in "if" test ?

Then what is next ? I do another pull request after have fixed your remarks ? it is my first time on github

mchamberland:
I don't know much about the dose actor or Geant4 radiation transport, but I disagree about the photon part. The code is clearly computing the stopping power for charged particles in order to calculate dose. Photons shouldn't be included there (if at all).

Have you tested your code before opening this pull request?

uhqd:
Dear Marc, First I had the same opinion. But yes, I tested it and observed that photons are also treated by this two sections (dose and dose to water). First I was surprised. I finally understand  that in GATE (after reading Sarrut et al MP 2014)  if energy is > electrons cut (I tested it with a huge value like 10 m) photons deposit dose at the "poststep position". If you set electrons cut at 10 m (dumb value), only photons are processed  in doseActor. You can test it by using example3 in example_Radiotherapy.

aelia21:
Hello!

I know that this is not the place for posting question, but I would actually like to understand what is the variable "gray" in the code stand for...? I cannot find its definition...

uhqd:
gray = 1e-12
In Geant4, units are in a consistant system (I suppose SI). If you want to display a particular unit you just divide by its name. It is not so obvious to find out 1e-12. You can play with it (i did it once) in G4SIUnits.hh. For example gray = joule / kilogram. And kilogram is 1 but  joule is kg*m*m/(s*s)....etc...

uhqd:
I did a new commit and a new pull request taking into account the remarks of Thomas. Last remark. In this code I don't understand why these 2 lines do not give the same result when material is water:
        DEDX = emcalc->ComputeTotalDEDX(100, "proton", material, cut);
        DEDX_Water = emcalc->ComputeTotalDEDX(100, "proton", "G4_WATER", cut);

I define my water like this 
Water: d=1.00 g/cm3; n=2 ; state=liquid
	+el: name=Hydrogen ; n=2
	+el: name=Oxygen; n=1

aelia21:
Thanks for your answer! 

Regarding your doubt: I think it is because there is a difference between water and G4_WATER in the Iw-value and therefore in the DEDX.

uhqd:
Thank you but what is lw-value ?

aelia21:
It is the water ionization potential. If you use Water as material  you could have a Iw of about 69eV. But if you use G4_WATER the Iw value is internally set to 78eV.

uhqd:
Ok thank you very much. Interesting because in my ICRU tables this value is ...75eV. I am wondering how the value 69eV is found

GrevillotL:
Hello,

In the new ICRU report, I=78eV.
I am actually thinking that the DoseToWater part could be more generic and be a DoseToWhatever, the Whatever could be given by the user (G4_WATER, water, but also any other material the user like to convert the dose to). But that is a further development….

Regarding the cut and “photon dose”  - As far as I understand, if the energy transferred by the photon is lower than the cut value, then the secondary electron will not be explicitly produced and the energy deposited on the spot by the photon. Nevertheless, the photon probably does not have a DEDX in Geant4 and I am not sure photon can simply be included in the list of particles – did you test that? It does not crash?.
That is why my question was to check if the particle is a photon, then maybe we should look for the DEDX of the electron at the corresponding energy. However, we test this hypothesis and check the interaction types and energy loss (I think at high energy there are also some photo/neutron interactions)

Best,
Loïc

uhqd:
really sorry. My previous commit+pull request got a small error. This one is ok

uhqd:
Hi again Loic
Thank you for your feed back.
First I would like to apologize again for my multiple commit... Realy realy sorry , I feel embarassed.
Concerning your point what I did was to just display the dose (and doseToWater) with the particle name. I was  expecting to see no photon or only photons but with  dose = 0 and doseToWater =0. 

But no... With the current git version, if you just add these two lines after corresponding dose computation. 

G4cout << "here we got a " << PartName << " dose  is " << dose << Gateendl;

and

G4cout << "here we got a " << PartName << " dose to water is " << doseToWater << Gateendl;

Here is a part of the output:
here we got a gamma dose to water is 8.44909e-12
here we got a gamma dose  is 5.40855e-13
here we got a gamma dose to water is 5.32651e-13
here we got a gamma dose  is 7.81739e-12
here we got a gamma dose to water is 7.69881e-12
here we got a gamma dose  is 1.09821e-12
here we got a gamma dose to water is 1.08156e-12
here we got a gamma dose  is 4.25923e-11
here we got a gamma dose to water is 4.19462e-11
here we got a e- dose  is 1.05237e-10
here we got a e- dose to water is 1.05265e-10
here we got a e- dose  is 1.01765e-11
here we got a e- dose to water is 1.01792e-11
here we got a e- dose  is 1.61474e-10


As you can see dose is deposited by electrons but also by gamma.
In my example, cut is 0,1 mm for both electrons and photons

uhqd:
Concerning the "I" value, I think I must update my ICRU books :-)
But  our clinical colleagues they feel uncomfortable with the fact that doseToWater is not exactely equal to Dose in a water phantom. 
I found the value 78eV for G4_Water but I still haven't found where it is computed for the "user water" (I mean 2 H + 1 O).

uhqd:
As i said on issue 104:
Finally  I found a simple way to obtain: dose = doseToWater in a water phantom
In my macro:
`/gate/patient/setMaterial G4_WATER`

instead of

`/gate/patient/setMaterial water`

source/digits_hits/src/GateDoseActor.cc
dsarrut: NO ! We must keep the dose in gray, this is the natural unit.

dsarrut:
Hello, 

regarding the issue raised by Luc during the last opengate meeting about the "dose_to_water" in the DoseActor. Could we agree on the modification to do ? 

It is around this line : https://github.com/OpenGATE/Gate/blob/develop/source/digits_hits/src/GateDoseActor.cc#L392

Proposition:
- 1) add || PartName== "gamma" ass suggested by Luc
- 2) remove the "else" part and replace by a user "warning" that the result will be false

Are you agree with that ? Other suggestion ? 

(PS: once all agree, I will made the change separately from this pull-request because we should not change the gray unit)

dsarrut:
Closed: discussion continue in #122

PR #108, opened 2017-03-23
TPS pencil beam: bug fixes and requested features

djboersma:
This pull request contains:
1. A bug fix for RT plan file reading. If a line would contain more words than expected (e.g. in the first content line: the plan name) you'd get a crash instead of the intended helpful error message. Now you do get this helpful message. After that Gate may still crash, but that is because ROOT/Geant4 fail to terminate gracefully.
2. As requested by MedAustron, the new vertex generation method can now also sample randomly from all spots. You get this when SortedSpotGenerationFlag is set to` false`.
3. As requested by MedAustron, the default interpretation of the energy spread polynomial is now relative (a percentage of the energy). The best place to explicitly configure is the source properties file, by putting the keyword 'MeV' or 'percent' in the line before the polynomial order, to get an absolute or relative interpretation, respectively.

I did some very rudimentary testing, using example5 in examples_Radiotherapy. I get the same X/Z/Ekine phase space distributions for spot-sorted and random generation. I ran with both absolute and relative energy spread, but the effect is too subtle to be noticed. Either that, or I did something wrong. Hopefully this will become clearer during MedAustron's testing & validation in the next few days/weeks.

The `OldStyleFlag` and the old vertex generation code is still present. If MedAustron's testing and validation gives us enough confidence that the new code is good, then it would be nice to delete the old code before we do the release.

### DIFF
--- a/examples/example_Radiotherapy/example5/data/Source-Properties.txt
+++ b/examples/example_Radiotherapy/example5/data/Source-Properties.txt
@@ -17,6 +17,7 @@
 
 # energy spread
 # polynomial order
+percent
 1
 # polynomial parameters (highest to lowest)
 0.01
--- a/examples/example_Radiotherapy/example5/mac/main.mac
+++ b/examples/example_Radiotherapy/example5/mac/main.mac
@@ -88,6 +88,8 @@
 #/gate/source/PBS/setNotAllowedFieldID 1
 #/gate/source/PBS/setNotAllowedFieldID 2
 /gate/source/PBS/setFlatGenerationFlag false
+/gate/source/PBS/setSortedSpotGenerationFlag false
+/gate/source/PBS/setSigmaEnergyInMeVFlag false
 /gate/source/PBS/setSourceDescriptionFile data/Source-Properties.txt
 
 #=====================================================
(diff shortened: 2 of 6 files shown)


djboersma:
Clarification on point 1: this bug was found because Alessio tried to use the code with a plan name that contained more than one word. I had somehow assumed that plan names would always be like that (my own RT plan files have single-word names) but apparently it is common to have more verbose names. So I changed this so that the name can be any length (but it should not be empty and it should be on a single line). I think it's still limited to ASCII characters, because that's what you get with `std::string`. In French/German/Scandinavian clinics users might be tempted to save their DICOM plan files with names that have éèàîöäåüøæ in them. If we ever decide that there is a need to support that, then we'll need to do update the parsing of the plan line. Should not be too hard, I think. (How does the clitk tool handle special characters?)

Lines that are not plan lines still have a hardcoded number of expected words, usually 1 or 3. A deviation from that will lead to an exception being thrown with the line number and filename to show the user where the parsing failed.

The parsing code is not fool proof, e.g. if on a particular line we expect to get one single integer, then you will get an error for "42 43" but not for "4.2 cm" (which will be parsed as 4). So the current parser is better than the old one, but it's not perfect. I do not want to spend my time writing completely perfect parsers. If we ever do want a perfect parser then we should maybe not try to write it ourselves but instead use an existing format, like XML or json, and let a library do the parsing.

dsarrut:
See  #117

PR #109, opened 2017-03-29
implements conditional compilation code for GATE_USE_DAVIS

wrzof:
commit c2003de40e3d4055d02d9a40f76e217baf7236c9 added some code which depends on modified Geant4 version even if GATE_USE_DAVIS is set to OFF.


albertine:
Thanks. It seems then I forgot to commit the proper GateSurface.cc file in my commit c2003de
My mistake, sorry...

PR #114, opened 2017-04-13
I have modified GateTessellated to read-in .stl CAD files.

MCvin:
It supports ASCII and binary .stl and both triangular and quadrangular facets.
STL format being the most common and widely used CAD format.

The orginal GateTessellated could only read-in ASCII files with a specific (non standard) format
containing three vertices per line and no header. It was implemented as an example.

Now STL files are supported!

### DIFF
--- a/source/geometry/include/GateTessellated.hh
+++ b/source/geometry/include/GateTessellated.hh
@@ -34,10 +34,13 @@ class GateTessellated : public GateVVolume
     // Declaration of pure virtual method
     inline virtual G4double GetHalfDimension( size_t ) { return 0.; }
 
-    void SetPathToVerticesFile( G4String );
+    void SetPathToSTLFile( G4String );
 
   private:
-    void AddFacet( G4ThreeVector, G4ThreeVector, G4ThreeVector );
+    void ReadSTL_ASCII();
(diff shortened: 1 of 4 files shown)


albertine:
Thanks Maxime !
Could you update the wiki documentation with some information regarding STL format ? (http://wiki.opengatecollaboration.org/index.php/Users_Guide_V8.0:Defining_a_geometry#How_to_build_a_.22tessellated.22_volume)
If you're OK, I can create an account for you on Mediawiki

MCvin:
Yes no problem, I will be happy to!

albertine:
Great ! 
I've just created your account. 
Enjoy

MCvin:
Thanks, I'll try!

albertine:
Last thing
Do you have an example that you would accept to make available to the Gate-users community ?
If yes, you can create a pull request  to the OpenGATE/GateContrib repository...

MCvin:
I was working on it ;)

MCvin:
I managed to log in to the wiki, I can edit but I cannot save my changes... the page loads and then fails.
Do you know why? Also how can I upload a file (.png) ?

PR #117, opened 2017-04-20
correct sigma energy, remove crashing debug msg

djboersma:
The choice between relative and absolute energy spread specification
should be used not only in the 'old' but also in the 'new' vertex
generation method. The default should be 'relative' (like in the docs)
but without this fix the value from sourceproperties would be used as
absolute (like it was in release 7.2).

I removed a debugging message which tried to use the value of an entry
in the NbProtonsToGenerate vector, but that vector has length 0 in case
the user opts for random spot sampling, which would result in an
embarrassing crash. We don't want that.


djboersma:
Alessio did his verification work at MedAustron with practically the same patch applied. This patch should have been committed to the `djboersma-TPS_fix` branch but I guess due to some communication glitch that never happened.

PR #122, opened 2017-05-22
DoseToWater 

dsarrut:
Hello, 

You will find here a pull request that try to take care of what have been said, being:
- add gamma in the if condition
- consider Electron DEDX if particle is a gamma
- ignore the 'else' condition of the loop
- replace ComputeTotalDEDX with pointer rather dans string comparison

Some remarks:

- when shooting proton, what about all the other generated particles: C13, O16, N14 (to only name a few), they don't directly contribute to the dose to water ? Before the modification, their deposited energy was added via 'proton' DEDX. Now, with this implementation, they are ignored. It seems wrong to me ?

- I implement the ComputeTotalDEDX with pointer, I think it will change anything about speed, but it is clearer, good catch. 

Could you please test and report ? 

Thanks,

### DIFF
--- a/source/digits_hits/src/GateDoseActor.cc
+++ b/source/digits_hits/src/GateDoseActor.cc
@@ -20,6 +20,11 @@
 #include <G4VoxelLimits.hh>
 #include <G4NistManager.hh>
 #include <G4PhysicalConstants.hh>
+#include <G4Gamma.hh>
+#include <G4Proton.hh>
+#include <G4Positron.hh>
+#include <G4Deuteron.hh>
+#include <G4Electron.hh>
 
 //-----------------------------------------------------------------------------
 GateDoseActor::GateDoseActor(G4String name, G4int depth):
@@ -46,6 +51,8 @@ GateDoseActor::GateDoseActor(G4String name, G4int depth):
 
   mVolumeFilter = "";
   mMaterialFilter = "";
+
+  mDose2WaterWarningFlag = true;
 
   pMessenger = new GateDoseActorMessenger(this);
   GateDebugMessageDec("Actor",4,"GateDoseActor() -- end\n");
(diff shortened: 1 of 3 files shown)


source/digits_hits/src/GateDoseActor.cc
```diff
@@ -379,29 +386,34 @@ void GateDoseActor::UserSteppingActionInVoxel(const int index, const G4Step* ste
...
-        //if (PartName != "O16[0.0]" && PartName != "alpha" && PartName != "Be7[0.0]" && PartName != "C12[0.0]"){
-
-        DEDX = emcalc->ComputeTotalDEDX(Energy, PartName, material, cut);
-        DEDX_Water = emcalc->ComputeTotalDEDX(Energy, PartName, "G4_WATER", cut);
-        doseToWater=edep/density/Volume/gray*(DEDX_Water/1.)/(DEDX/(density*e_SI));
-        if (DEDX_Water == 0 || DEDX == 0) doseToWater = 0.0; // to avoid inf or NaN
+      double volume = mDoseToWaterImage.GetVoxelVolume();
+
+      // Get current particle
+      const G4ParticleDefinition * p = step->GetTrack()->GetParticleDefinition();
```
erialm: Get rid of the const here, I think. Line 406 will cause problems trying to assign a new value to a const.

dsarrut: Not sure ... I think it is a pointer to const and not a const pointer, meaning the object pointed to is const.

djboersma: I agree with @dsarrut here. :)

GitFuchs:
Is there a reason why DEDX is not calculated for every particle (performance etc) ?
Like: If not a photon do...

Issue #94, opened 2016-10-07
Gjs parsing issue with multiple, same aliases per line


canonnervio:
Apparantly gjs has an issue parsing the same alias more than once per line, e.g. a command like 

`/gate/SPECThead/placement/setTranslation       {transl} {transl} 0 mm`

with the parameter 'transl' e.g. '5' passed to it is parsed to 

`/gate/SPECThead/placement/setTranslation       5 {transl} 0 mm`

instead of

`/gate/SPECThead/placement/setTranslation       5 5 0 mm`

djboersma:
Hi @canonnervio, could you try replacing `if` by `while` in line 224 of `cluster_tools/jobsplitter/src/GateMacfileParser.cc`? If that solves the problem: could you submit a pull request with it? :)

PR #131, opened 2017-06-26
fixed Gjs parsing issue with multiple, same aliases per line #94

canonnervio:


canonnervio:
Fixed [this](https://github.com/OpenGATE/Gate/issues/94) issue as suggested by @djboersma.

PR #134, opened 2017-07-13
Rename CUDA compute_21 to compute_20; there is no compute_21!

AndiH:
In commit 2c94cf6b79f9d0d7a5b2b0ca29dff0228a3674c2 @ekleung expanded the list of target Compute Capabilities of GPU devices. My guess is, they did so with copy and paste. There is no `compute_21` (although there is a `sm_21`, confusingly). This commit fixes it.

This was spotted by @djboersma in December (https://github.com/OpenGATE/Gate/commit/2c94cf6b79f9d0d7a5b2b0ca29dff0228a3674c2#commitcomment-20230695) but never fixed.

@djboersma also mentions backwards-compatibility of CUDA versions. One suggestion from my side would be to introduce some guarding in the `CMakeLists.txt` for it. Like in Gromacs: https://github.com/gromacs/gromacs/blob/master/cmake/gmxManageNvccConfig.cmake#L121-L135. This could also cover the deprecation for Compute Capability 2.0 in CUDA 9 and beyond (like `if CUDA < 9.0: add sm_20; else: don't`).  
If you're interested in that, let me know, I can work on that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,7 @@ ENDIF()
 #=========================================================
 # Add the executable, and link it to the Geant4/ROOT/CLHEP/ITK libraries
 IF(GATE_USE_GPU AND CUDA_FOUND)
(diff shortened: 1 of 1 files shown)


xiezhaoheng:
**Thanks for you guys' help!!!**, we solved the problem of  **"​nvcc fatal: unsupported GPU architecture 'compute_21'** "~
Just a quick reminder to the quite new users: **"rename the number, from "compute_21" to "compute_20", in the 228th line of  'Updated CMakeLists.txt'**.  
By the way, we successfully installed  GATE8.0 with GPU by using CUDA 8.0, GTX660 and ubuntu 16.04.

'

PR #142, opened 2017-09-20
Mandatory C++11

### LINKED ISSUE #90: How long will we keep C++11 optional?
I wonder when we should start requiring developers/users to compile  Gate with C++11. This "new" standard is now 5 years old, there is already a newer one (C++14) and in 2017 there will be yet another one. I am not necessarily advocating that we should jump to the bleeding edge, but I do think that at some point we should decide that we do not need to keep our code 2003-compatible anymore. All the tools that we depend on (geant4, root, ITK) have provided C++11 versions already long ago.

I think that the main argument for keeping the code 2003-compatible is to be nice to users with very old systems, running Ubuntu 12/04 or even 10/04. But I think/hope that those users are getting very rare, and maybe some of them can be gently encouraged to join the modern times with a shiny new linux distro.

On the other hand, if we do not want to repeat this discussion every 3 years (because on https://isocpp.org/std/status you can see that after 2017 the standards committee won't stop innovating) then some of us might even be tempted to leapfrog directly to c++2014. I am reading Scott Meyers' "Effective Modern C++" and I get the feeling that C++14 is kind of a "bugfix release", making many of the new elements in C++11 more convenient and consistent. But you can't compile C++14 code with gcc 4.8, which I think is still very popular.

I am opening this issue because I think that there needs to be a discussion and eventually a smart & practical policy/decision/agreement/compromise about our choice of C++ standard(s). I guess we can collect facts & opinions & things to consider, and then discuss it all in Strasbourg.

Supernabla:
Fixes #90

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ GTAGS
 *bidon*
 *docx
 *save*
+.vscode
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,6 @@ SET(GATE_USE_ROOT ON)
 SET(GATE_ANALYSIS_USE_GENERAL ON)
 SET(GATE_ANALYSIS_USE_FILE ON)
 
-OPTION(GATE_USE_STDC11 "Use the c++11 standard in the compilation" ON)
 OPTION(GATE_USE_OPTICAL "Gate use optical photon" OFF)
 OPTION(GATE_USE_DAVIS "Gate use Davis LUT model" OFF)
 OPTION(GATE_USE_LMF "Gate use LMF" OFF)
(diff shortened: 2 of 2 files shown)


djboersma:
Yay! 👍 

What is "vs-code" for? Are you building GATE in Visual Studio on Windows?

Supernabla:
Recently, I've been testing "Visual Studio Code", which is Microsoft's Open Source pendant to MS Visual Studio. It actually runs on Linux (I'm using Ubuntu 16.04). Like other IDEs, it creates a project folder, i.e. ".vscode", where it stores debugger/parser/editor settings. Git needn't track that.

mojca:
Some time ago I spent a considerable amount of time trying to get Gate working on Windows and nearly succeeded, but some further patches would be needed. It's not entirely hopeless if someone invests a bit of time, all dependencies already work on Windows.

Supernabla:
To be quite honest, I didn't think about anything related to Windows, when suggesting these commits. I just want to propose to get rid of `GATE_USE_STD_C11`, because it's an obsolete cmake variable. My addedum to .gitignore is rather irrelevant, but practical.

dsarrut:
thanks !

PR #145, opened 2017-10-02
BUG: Fix output of volume names in binary files

### LINKED ISSUE #141: Strings Broken in Binary Format
In the binary implementation of the output in [GateToBinary.cc](https://github.com/OpenGATE/Gate/blob/89a3b77f801b2c98cf24609306639b0d9fde07d7/source/digits_hits/src/GateToBinary.cc#L265-L270) (the hits output listed at that link) improperly writes out the G4String structure, rather than the contents of the string.  Below is the code used for Singles output on Line 825.

    m_outputFile.write( reinterpret_cast< char* >( &compVolName ), sizeof( G4String ) );

Using vGate, which I think used GCC (<v5.0) to compile on a 64bit system, the value written out 8 bytes of garbage.  [G4String](https://github.com/Geant4/geant4/blob/1a1316fea4d33260fe6fba9e5ece5149b9c877d0/source/global/management/include/G4String.hh#L59) in Geant (improperly) derives from std::string.  The implementation of std::string is compiler std library dependent, which means the sizeof(G4String) will differ between 32bit and 64bit platforms and different compliers.  The value, though, is wrong on every platform.

The proper format would be to chose a fixed number of characters to print so that each hit or single event is the same size in bytes, or to write out the string value like something this:

    m_outputFile.write(compVolName.c_str(), compVolName.length() + 1);

This is using the std::string interface, not the G4String interface.

I would prefer the fixed size that truncated after a point.  Currently the only way to not break the binary output for GATE compiled with GCC <v5.0 on 64 bit systems would be to truncate to 7 characters or fewer to have a null terminated string, or 8 if you would want to ignore that.  Anything else will break the current binary format.

Any preferences on how to fix this?

dfreese:
In the binary output format, G4String were being improperly written out so that
the field had invalid data.  This issue is fixed by writing out the volume
string up to a fixed length.  If shorter, the rest of the field is filed with
zeros.

fixes #141

### DIFF
--- a/source/digits_hits/include/GateToBinary.hh
+++ b/source/digits_hits/include/GateToBinary.hh
@@ -398,6 +398,9 @@ protected:
 
 	std::ofstream m_outFileRun; /*!< outfile for run */
   std::ofstream m_outFileHits; /*!< outfile for hits */
+
+private:
+    static G4String FixedWidthZeroPaddedString(const G4String & full, size_t length);
 };
 
 #endif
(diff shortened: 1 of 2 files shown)


dsarrut:
thanks !

PR #149, opened 2017-10-29
Commented out /gate/physics/addProcess Fictitious

ekleung:
When /gate/physics/addProcess Fictitious is enabled, Geant4 will produce an error that will either cause GATE to either crash or get stuck. In my case, it was stuck in GATE v8.0 and crashed in GATE v7.2 develop (pull request #66).

Below is an example of the error:

*** G4Exception : GeomNav0003
      issued by : G4PhantomParameterisation::GetReplicaNo()
Point outside voxels!
        localPoint - (-nan,-nan,-nan) - is outside container solid: mat25_solid

*** This is just a warning message. ***

Commenting out the command fixed the crash, although I do not know if it will affect more than just computational time. Perhaps the command is simply broken in the newer versions of GATE? From what I understand, the example does use the interfile format already:

http://wiki.opengatecollaboration.org/index.php/Users_Guide_V8.0:Voxelized_Source_and_Phantom#Fictitious_interaction

On a side note, the example is 4+ years old. GATE wiki currently recommends the use of voxelized phantom descriptions such as ImageRegularParametrisedVolume instead of fictitiousVoxelMap as provided in the example. An update of the example would be very helpful for beginners like myself. I will play around with the example to see if I can get it running with the "new" commands but if someone who knows this stuff can quickly provide an updated example it would be highly appreciated :)

### DIFF
--- a/benchmarks/benchVoxelRegular25x25x25/physicsFGATE.mac
+++ b/benchmarks/benchVoxelRegular25x25x25/physicsFGATE.mac
@@ -1,7 +1,7 @@
 #############
 #  Physics  #
 #############
-/gate/physics/addProcess Fictitious
+#/gate/physics/addProcess Fictitious ### BROKEN AS OF 20171029 - DO NOT USE ###
 
 /gate/physics/addProcess PhotoElectric
 /gate/physics/addProcess Compton


aelia21:
Hello!

I am not an expert on this side but I would like to ask something. As far as I remember the command addProcess(es) were not working anymore since the physics BUILDER approach was introduced (Gate 7.x I guess)... am I wrong? 
Also, are we sure that to comment part of the code out would be the best solution?  

Cheers!
Alessio

ekleung:
Hi Alessio,

I've been using the command /gate/physics/addProcess [something] without noticing any issues in both GATE v8.0 and some v7.x versions. If I recall correctly, the physics builders may not include processes such as RadioactiveDecay. I do not have a good understanding of which physics builder is best for my application so I continued using the addProcess method.

Simply commenting out the command is certainly not the best solution but it stops GATE v7.2 and v8.0 from crashing in my case. I did not want to delete the entire line until official GATE devs say it's okay to delete that line (in that case they'll be able to edit my file anyway). I hope other devs will chime in on this.

Thanks,

Ed

dsarrut:
I think Fictitious no longer work and must be disabled, but we need confirmation from people that work with it before removing from the source code.

djboersma:
What do we do with this, should this be included in the 8.1 release? If that line of code only causes crashes then I am in favor of commenting it out (or better: deleting it). 
@dsarrut , who are "the people that work with it" and did you succeed in contacting them?

PR #157, opened 2017-11-28
fixes division-by-zero error in linear interpolated user energy spectrum

### LINKED ISSUE #129: Bug - linear interpolated spectrum - radionuclide energy spectrum
Hi

I would like to report a Bug.
I am using Y-90 source and I change the spectrum from MIRD database.
In one part of the Y-90 beta spectrum there is two point with these values:

E(MeV) Probability
0.70000 6.274E-01
0.75000 6.292E-01
0.80000 6.292E-01
0.85000 6.276E-01

In my simulation the energy spectrum in the region between 0,75MeV and 0,80MeV, just fall down. 

With the help of other Gate-Users we found that the problem come from the interpolation between this two point (( .75000, 6.292E-01),(0.80000. 6.292E-01)). I don´t know the procedure that Gate follow to do the interpolation, but this is an error. 
When I changed slightly for example:
0.70000 6.274E-01
0.75000 6.292E-01
0.80000 6.291E-01
0.85000 6.276E-01

The error disappear. 

Now I am using  histogram spectrum instead of linear interpolated spectrum.
With histogram spectrum the error is not present.

Supernabla:
Fixes #129

### DIFF
--- a/source/physics/include/GateSPSEneDistribution.hh
+++ b/source/physics/include/GateSPSEneDistribution.hh
@@ -23,48 +23,48 @@ See GATE/LICENSE.txt for further details
 #ifndef GateSPSEneDistribution_h
 #define GateSPSEneDistribution_h 1
 
-#include "G4PhysicsOrderedFreeVector.hh"
-#include "G4ParticleMomentum.hh"
-#include "G4ParticleDefinition.hh"
-#include "G4DataInterpolation.hh"
+#include <vector>
 
-#include "G4SPSEneDistribution.hh"
+#include <G4Types.hh>
+#include <G4String.hh>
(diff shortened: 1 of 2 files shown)


Supernabla:
Although not related to the referenced issue, I changed all `G4double*` to `std::vector<G4double>` in the `GateSPSEneDistribution` class. To my understanding this will **NOT** slow down the call of `GateSPSEneDistribution::GenerateFromUserSpectrum()` (which is called quite often during a run...).

The random access `operator[]` of a `std::vector` should be just as fast as calling from a c-array, given that Gate is compiled with `-O3` (i.e. `Release` configuration).

PR #158, opened 2017-12-01
Reduced memory foootprint of voxellized sources

patayg:
The current implementation of voxellized sources uses the following data representation:
`typedef std::map<std::vector<G4int>,G4double>   GateSourceActivityMap;`
or 3 integers + 1 double for _each_ voxel (on my architecture, that's 20 byte).
The proposed modification uses the following representation:
`typedef std::vector<G4double> GateSourceActivityMap;`
or 1 double for each voxel (12 byte less than the previous), and we can use some basic modulo arithmetic on the vector's index to retrieve the voxel 3-indices.

**Pros:**
- less than half the memory consumption
- easier to read the code
- no need to manually re-convert into linear format (as seen in `GateGPUIO_Input_parse_activities`)

**Cons:**
- a possible minimal computational overhead (inline modulo operations vs. retrieving the components of a 3-vector)
- cannot handle arbitrarily placed cube-clouds generated by GateSourceVoxelTestReader (is there any reason to support those anyway?)

I removed a lot of range-checks because the linear indexing scheme inherently ensures valid index values.


dsarrut:
thanks !

PR #160, opened 2017-12-05
Tetrahedral-Mesh Geometry

Supernabla:
The first commits contain a new `GateVVolume` descendent, the `GateTetMeshBox`, which is able to load tetrahedral-mesh geometries from ASCII files (`.ele`) which can be produced using a tool called [TetGen](www.tetgen.org). Here, some visualisations from GATE simulations I ran:

tetmeshbox

tetmeshboxrun

----

The last commit introduces the `GateTetMeshDoseActor`, an actor, that can only be attached to `GateTetMeshBox` volumes. This is an interface similar to the `GateDoseActor`. It scores dose (and rel. uncertainty) deposited in each tetrahedron. The output file is a simple csv-table:

```
# Tetrahedron-ID, Dose [Gy], Relative Uncertainty, Sum of Squared Dose [Gy^2], Volume [cm^3], Density [g / cm^3], Region Marker
0, 1.3e-08, 1.3e-01, 3.0e-18, 1.9e-02, 9.5e-01, 1
1, ...
...
```

### DIFF
--- a/source/digits_hits/include/GateTetMeshDoseActor.hh
+++ b/source/digits_hits/include/GateTetMeshDoseActor.hh
@@ -1,0 +1,84 @@
+/*----------------------
+  Copyright (C): OpenGATE Collaboration
+
+  This software is distributed under the terms
+  of the GNU Lesser General  Public Licence (LGPL)
+  See LICENSE.md for further details
+  ----------------------*/
+#ifndef GATE_TETMESH_DOSE_ACTOR_HH
+#define GATE_TETMESH_DOSE_ACTOR_HH 
+
+#include <memory>
+#include <map>
+
+#include <G4Types.hh>
+#include <G4String.hh>
+#include <G4THitsMap.hh>
+#include <G4HCofThisEvent.hh>
+#include <G4Step.hh>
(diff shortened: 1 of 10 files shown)


Supernabla:
There is a [corresponding pull request](https://github.com/OpenGATE/GateContrib/pull/16) to the [GateContrib](https://github.com/OpenGATE/GateContrib) repository, which contains an example to test the code.

dsarrut:
thanks a lot ! 

Could you please add few lines in the doc in the wiki : http://wiki.opengatecollaboration.org/index.php/Users_Guide_V8.1

Also, I have the following warning : 
`source/digits_hits/src/GateTetMeshDoseActor.cc:131:5: warning: designated initializers are a C99 feature [-Wc99-extensions]`
Could you try to get rid of it ? 

thanks
David

yuxiXo:
Dear Supernabla，
May I ask you a question?
How to use the new meshed approach for imaging applications. For example, the imaging of  “TetMeshBox” volume, and get output of projection set.
Is it possible to achieve the function？

thanks
Xixi Yu

PR #161, opened 2017-12-07
replaced designated initialisers with aggregate initialisation

Supernabla:
@dsarrut concerning the compiler warnings you had


dsarrut:
perfect, thanks ! (even if it is a bit less clear in the code, at least it didn't complain now during compilation). 
Your contribution is really appreciated !

PR #164, opened 2018-01-18
Update CREDITS.txt

uhqd:


uhqd:
Uwe is the new spokesperson

dsarrut:
CREDITS.txt no more exist (see AUTHORS)

PR #168, opened 2018-03-06
use track length in voxel in xsection calc

djboersma:
This is work done by Heide Meissner, a TU Wien student working for/with Hermann Fuchs.
Hermann gave me this code to merge into github.

The main change is to store the total track length inside every voxel and use
that to replace the factor "voxel volume / beam cross section" in the formula
that computes the interesting observable. According to the comments this
enables running the actor for beams under any angle. I suppose that secondary
particles were already running under "any angle" so also for beams parallel to
an axis this improved calculation could be worthwhile.

### DIFF
--- a/source/digits_hits/include/GateCrossSectionProductionActor.hh
+++ b/source/digits_hits/include/GateCrossSectionProductionActor.hh
@@ -11,9 +11,14 @@ See LICENSE.md for further details
   \class  GateDoseActor
   \author claire.vanngocty@gmail.com
  */
+
+/* Modified: replaced length of voxel by path length of proton in voxel. This enables arbitrary beam directions. Author: Heide Meissner, TU Wien */
+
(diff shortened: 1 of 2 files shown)


djboersma:
I edited Heide's files a little bit so that this PR would not clobber the cosmetic fixes that have been done to these two source files in the mean time (Heide worked on this quite a while, if I understood correctly). However, I see that there is some code convention work to do: lots of tabs and weird white space stuff. Did not want to mix that into this commit.

djboersma:
Who is using this cross section actor? I did not find any uses of this actor in GateContrib, so it's not so easy to test this new code. Could somebody upload a CrossSection example? Or better even: add a benchmark test for it? :)

It looks like @sj202988 is the original author of this actor. Could he/she please comment on this PR? Should we include this in the upcoming 8.1 release or leave it for later?

Issue #123, opened 2017-05-25
itk-mhd installs files (to the wrong place)


mojca:
I'm experiencing this problem with Gate 8.0.

I'm using different directories for:
- Gate sources
- build directory
- installation directory
- destroot directory (for `make install DESTROOT=/temporary/prefix`)

I do quite a bit of manual shuffling of the files around (creating a shell script that defines the files from Geant4, putting materials database and macro files to the right place etc.). But there is one weird aspect of running `make install DESTROOT=/temporary/prefix`. I end up with the following files:
```
./opt/local/bin/Gate
./opt/local/include/localMetaConfiguration.h
./opt/local/include/metaArray.h
./opt/local/include/metaArrow.h
./opt/local/include/metaBlob.h
./opt/local/include/metaCommand.h
./opt/local/include/metaContour.h
./opt/local/include/metaDTITube.h
./opt/local/include/metaEllipse.h
./opt/local/include/metaEvent.h
./opt/local/include/metaFEMObject.h
./opt/local/include/metaForm.h
./opt/local/include/metaGaussian.h
./opt/local/include/metaGroup.h
./opt/local/include/metaImage.h
./opt/local/include/metaImageTypes.h
./opt/local/include/metaImageUtils.h
./opt/local/include/metaIOConfig.h
./opt/local/include/metaITKUtils.h
./opt/local/include/metaLandmark.h
./opt/local/include/metaLine.h
./opt/local/include/metaMesh.h
./opt/local/include/metaObject.h
./opt/local/include/metaOutput.h
./opt/local/include/metaScene.h
./opt/local/include/metaSurface.h
./opt/local/include/metaTransform.h
./opt/local/include/metaTube.h
./opt/local/include/metaTubeGraph.h
./opt/local/include/metaTypes.h
./opt/local/include/metaUtils.h
./opt/local/include/metaVesselTube.h
./opt/local/lib/libMetaIO.a
./some/weird/build/path/gate/work/build/itk-mhd/libitkzlib.a
./some/weird/build/path/gate/work/build/itk-mhd/Utilities/itkzlib/itk_zlib_mangle.h
./some/weird/build/path/gate/work/build/itk-mhd/Utilities/itkzlib/zconf.h
./some/weird/build/path/gate/work/build/itk-mhd/Utilities/itkzlib/zlib.h
./some/weird/build/path/gate/work/build/itk-mhd/Utilities/itkzlib/zlibDllConfig.h
```

[...]

Issue #75, opened 2016-06-06
policy for the install directory


djboersma:
When running `make install` in a Gate build directory, the following things get copied to the install directory:

1. the (statically linked) Gate executable
2. the MetaIO library (static)
3. the MetaIO header files

Why do we install the MetaIO stuff? It's not even our own code, it's a copy of a directory in an old release of ITK. The static library is (by definition) not needed for running Gate. I don't think we want to encourage other developers to use our ITK copy (library and header files), they should use ITK directly.

On the other hand, I think it would be very useful to store the examples stuff (macros and input data) in the `share` part of the installation directory (except if the user left `GATE_DOWNLOAD_EXAMPLES_DATA` as OFF, then the user apparently does not want to install the examples). I usually treat the build directory as a temporary directory, which I delete after I'm done for the day with installing/developing/testing. All the interesting long term useful stuff should go into the install directory, and this includes the examples, in my opinion.

Another thing which I find myself doing all the time, is to create an `gate_env.sh` script, which I copy into the `$(CMAKE_INSTALL_PREFIX)/bin` directory. This file can be sourced and this then makes sure that my environment contains all the ROOT and Geant4 stuff that was used for this particular installation. Plus of course the bin directory is prepended to PATH. This is similar to the `thisroot.sh` and `geant4.sh` scripts that you get with ROOT and Geant4 (and of course my script uses those other scripts).

tl;dr: I want to change the install dir policy as follows:
* remove the MetaIO stoff
* add the examples (IFF the users enabled example data download)
* add a script for setting the environment of this particular Gate install

BishopWolf:
I dont think MetaIO stuff can be removed since ITK sources are not available as part of any distribution. So we must bundle it. The problem is what we bundle is an old copy of MetaIO and that can be updated.

djboersma:
These are two very different things:
(1) Including (a copy of) MetaIO source code in the Gate source code
(2) installing MetaIO as a compiled library & header files in the Gate install directories 

Your comment is about (1), while this issue is about (2).  Afer `make install`, I can delete the build directory as well as the `lib` and `include` subdirectories from the install directory, and `Gate` still runs. We need `MetaIO` only to build `Gate`, not to run it.

Maybe you guys never run `make install`, do instead everything always in the build directory?

SimonRit:
I agree with David, all 3 changes are good suggestions and I'd support a
pull-request in this direction.
Simon

djboersma:
[...]

PR #169, opened 2018-03-07
install only the Gate binary, no (external) headerfiles/libraries

djboersma:
Since ages the 'make install' command has installed not only the Gate binary,
but also some random stuff from the `itk-mhd` external source directory. With this
commit that should be fixed: 'make install' will *only* install the Gate binary
in the `$CMAKE_INSTALL_PREFIX/bin` directory, nothing else.

To achieve this, I commented something out in one `CMakeLists.txt` file and set a
flag in another. This way the offending code that was
responsible for these spurious installs is still there, but it is deactivated.
If someone remembers or stumbles upon a use case in which these lines were
actually useful, then we can re-enable them with the proper conditionals. If
this does not happen within a year or so, then we can just delete this dead
code.

This commit should fix issues #75 and #123.


djboersma:
@mojca , could you please check if this indeed fixes issue #123 that you reported? I checked 'make install', but I have no experience with "DESTROOT".

mojca:
Million times better, thank you very much. But please reword the commit message before merging.

djboersma:
sorry about that crude message. hope it's better now.

mojca:
You need to change the commit message rather than title of this ticket

djboersma:
I don't see a way to do edit that. I think I'll just delete this PR and resubmit.

mojca:
`git commit --amend` and edit the message
`git push -f <wherever you push>`

djboersma:
Cool, thanks! I think that worked. Are you OK with the new phrasing?

mojca:
Certainly ok by me, thanks a lot.

PR #170, opened 2018-03-09
X ray phase contrast

SimonRit:
The intention of this code is two-fold:
- To implement x-ray refraction, see http://wiki.opengatecollaboration.org/index.php/Users_Guide:Setting_up_the_physics#X-ray_at_boundary
- To implement Fresnel phase contrast imaging, see http://wiki.opengatecollaboration.org/index.php/Users_Guide:Tools_to_Interact_with_the_Simulation_:_Actors#Fixed_Forced_Detection_CT_with_Fresnel_phase_contrast
To this end, an (optional) dependency to tschoonj/xraylib has been added. Fresnel imaging is based on fixed force detection and, therefore, also requires a dependency to SimonRit/RTK.
Examples are provided in OpenGATE/GateContrib#17 and OpenGATE/GateContrib#18.

Two additional features for FFT are included in this PR:
- Parallel imaging for FFD but this requires to have compiled RTK with SimonRit/RTK@1325b71.
- Binning of FFD output.


djboersma:
Short:
I think that the `cmake` error message for not finding `xraylib` needs to be clarified/corrected: adding a directory to `PATH` did not solve it for me. This is a problem for users who cannot install `xraylib` in `/usr/local`.

Long:
I tried this out. I installed `xraylib` in a non-standard directory (with `./configure --prefix=/home/boersma/Software/Installations/XRAYLIB`). Then `cmake` tells me `"xraylib not found! Please install xraylib and make it available in PATH"`. So I added `/home/boersma/Software/Installations/XRAYLIB/bin` to my PATH variable. That did not help at all. Adding ``/home/boersma/Software/Installations/XRAYLIB/lib64` to my `LD_LIBRARY_PATH` did not help either. I "solved" it eventually with this incantation:
```
ccmake \
-DLIBXRL_LIBRARY:FILEPATH=/home/boersma/Software/Installation/XRAYLIB/lib64/libxrl.a \
-DLIBXRL_INCLUDE_DIR:PATH=/home/boersma/Software/Installation/XRAYLIB/include/xraylib \
../../Sources/GATE/Gate
```

This made `cmake` happy and allowed a successful build of Gate. After adding the RTK library path to my LD_LIBRARY_PATH I could also run the `imaging/CT/XRayRefraction` example in `GateContrib`.

So I wonder what we should recommend to the users. Maybe urge them to install xraylib in the standard location. But users do not always have access to `/usr/local`, e.g. when they run their stuff on a university cluster, on which they need to install their own special stuff on a user directory, like above. Is there a more advanced/subtle way to guide "pkgconfig" to find my xraylib installation? If yes, maybe the cmake error message could include a hint to that?

SimonRit:
I can't reproduce the problem on my MacOS. I don't know why but I have updated the error message.

PR #180, opened 2018-03-27
 G4EmProcessOptions class is replaced by G4EmParameters

gsize:
G4EmProcessOptions class is obsolete and will be removed in the next public release of Geant4


dsarrut:
thanks a lot !

PR #198, opened 2018-10-22
In the Dose By Region output, the number of hits and event_hits were …

MCvin:
…+1 wrong because of the last update method invoked to finish the squared dose calculation. I have corrected the output with nb_hits-1 and nb_event_hits-1.


dsarrut:
:+1:

PR #199, opened 2018-10-25
Multiple additions

TomThePhysicist:
This pull request contains several additions:  
- Optimization of the VoxelizedMass class memory usage
- Integration of the VoxelizedMass methods into the KermaActor
- Addition of a method to generate a mass image of a voxelized volume
_Usage :_  
`/gate/{volume name}/geometry/buildAndDumpMassImage {path to mass image + mass image filename (.mhd)}`
- Addition of a method to generate a voxelized image from a arbitrary unvoxelized geometry
_Usage :_  **(it is mandatory to use this method on the world volume)**
`/gate/world/setSaveImageDirectory  {path to a folder (image and HU2mat filenames are automaticaly generated)}`  
`/gate/world/saveGeometryAsVoxelizedImage {x voxel size} {y voxel size} {z voxel size} {unit (mm)}`


djboersma:
Cosmetic wish: rename "dump" to something more postive. In daily speech, "dump" has to do with negative things, like "toxic waste dump" or "she dumped her bad boyfriend". Also in computing, "dump" means usually something unpleasant, like in "core dump". But your new feature is something useful and nice, so the word "dump" seems out of place. So how about renaming `setDumpPath` to `setSaveImageDirectory`, and `dumpVoxelizedGeometry` to `saveGeometryAsVoxelizedImage`, for example?

TomThePhysicist:
Good remarks, I have changed the name of the commands as you suggested.

PR #201, opened 2018-11-02
dump the voxelized activity image

gsize:
dump the voxelized activity image, easy to check the activity distrivution


djboersma:
Same "cosmetic" comment as for #199 , would you maybe rename everything that contains 'Dump' to something more neutral but equally descriptive? Why not simply "Save"?

PR #202, opened 2018-11-06
Export the voxelized activity image

gsize:
After import the voxelized source,  can export an image to see the distribution of activity by imageJ:
 commands : 
`/gate/source/<source_name>/imageReader/readFile  <your_voxelized_source.mhd>`
`/gate/source/<source_name>/imageReader/exportActivityImage <export_image.mhd>`


dsarrut:
very useful, thanks !

PR #205, opened 2018-11-19
PSF time now double (was float). Both write (Actor) and read (Source)

dsarrut:
GatePhaseSpaceActor used to compute & write the particle time in float. Sometimes not enough, changed to double. 

GateSourcePhaseSpace detect if stored time is float or double.


djboersma:
I think this is fine. I remember we had some discussion about this about half a year ago. But I cannot find it, neither in the mail archive nor in the GitHub issues. It would be good to send a message to the user who originally raised this issue, that we think it's now solved. In order not to forget about this, I am waiting with merging and closing this request until we find that original discussion back... :)

PR #212, opened 2019-01-21
added fluence calculation EnergySpectrum

andiresch:
changed:
- by default not all spectra are generated
- new: fluence calculation from the track length and 1/cos(dz) approximation. The latter is only valid for certain beam angles.


andiresch:
yes, just updated wiki for energySpectrumActor

Am Do., 24. Jan. 2019 um 10:26 Uhr schrieb David Sarrut <
notifications@github.com>:

PR #219, opened 2019-02-01
add option to allow out-of-bounds HU values

djboersma:
Please do not merge this until the questions below are answered.

This patch was made upon user request: CT images often have small
number of voxels with HU values that are way out of the range of the
HLUT, due to image reconstruction artifacts or small metal elements
in the couch. The user would like that GATE automatically clamps the
HU values to the range of the HLUT table, i.e. HU values that are too
large are mapped down to the maximum HU value (or "last" material
in the table) and HU values that are too low are mapped up to the
material with label 0. The patch allows this, if the user provides
a positive value for the fraction of the voxels that is allowed
to be "too large". The idea is that the user should decide whether
the anomalous voxels are irrelevant or that the HU table needs to
be extended.

While implementing this fix I discovered two things that I did not
expect:

1. In line 348-351 of the original code there are two lines with
the comment "FIXME: remove these two lines". So I removed these two
lines. They add a material to the table for the "outside value".
Apparently this is controversial, so I removed this.
2. For CT images loaded as `ImageNestedParametrisedVolume`, the
`LoadImage(bool)` is called with the value `false`, which means that
no extra layer of voxels is added. For Regionalized images, `LoadImage`
is called with `true`. I'm not sure how to deal with that case.

If the "outside" of an image is air or vacuum, then using the lowest
material label for "outside" voxels is probably not very wrong.
But I guess it could also be water? Then it won't work.

Questions:
1. is it OK to remove those two lines flagged with the FIXME comment?
2. how should we deal with the "extra layer" of voxels?

### DIFF
--- a/source/geometry/include/GateVImageVolume.hh
+++ b/source/geometry/include/GateVImageVolume.hh
@@ -165,6 +165,7 @@ public:
   void SetDensityImageFilename(G4String filename);
   void SetMassImageFilename   (G4String filename) {mMassImageFilename = filename;}
   void EnableBoundingBoxOnly(bool b);
+  void SetMaxOutOfRangeFraction(double f);
 
 protected:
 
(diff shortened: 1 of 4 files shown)


dsarrut:
Sorry I push "merge" to quickly ;) David: could you correct the two lines about "outside value' as we discussed before ;) ? thanks !

Issue #203, opened 2018-11-09
Error when executing a visu.mac before intialization


MCvin:
I don't know if this is new but when the /control/execute visu.mac is done early in the main.mac this error occur:
[Core-0] Initialization of physics
[Core-0] Initialization of actors
GateImageT.icc (l.75): Unknown image file extension. Supported extensions are : 
.vox, .hdr, .img, .mhd, .mha, .h33, .i33, .dcm

This happens even when there is no image used in the macro so the error message is unlikely correct.

If the visu.mac is executed at the end, just before the application start. everything runs fine.

The problem seems to appear for any application with GATE 8.1.p01

acuihua:
can i have a look at your files? and may find the error

MCvin:
For example the error occur with this GateContrib example:
https://github.com/OpenGATE/GateContrib/blob/master/dosimetry/Radiotherapy/example2/
when running the macro file main-NPV1.mac with the visu (line 16 un-commented):
https://github.com/OpenGATE/GateContrib/blob/b7e3d0be41d0a2db5a94204899e65cb634d4d9b2/dosimetry/Radiotherapy/example2/mac/main-NPV1.mac#L16

The error has some randomness that I cannot explain. Sometimes it executes, sometimes it fails with the aforementioned error.

dsarrut:
See #222

PR #222, opened 2019-02-12
Fix issue #203 - Dose by region flag was not initialized.

SimonRit:
I used to valgrind to fix this. mDoseByRegionsFlag was not initialized so issue #203 was randomly occuring. With mDoseByRegionsFlag set to true, I had the same error message. I added an error message to make the error more explicit but set mDoseByRegionsFlag to false by default.

### DIFF
--- a/source/digits_hits/src/GateDoseActor.cc
+++ b/source/digits_hits/src/GateDoseActor.cc
@@ -68,6 +68,7 @@ GateDoseActor::GateDoseActor(G4String name, G4int depth):
   mVolumeFilter = "";
   mMaterialFilter = "";
   mTestFlag = false;
+  mDoseByRegionsFlag = false;
 
   pMessenger = new GateDoseActorMessenger(this);
   GateDebugMessageDec("Actor",4,"GateDoseActor() -- end\n");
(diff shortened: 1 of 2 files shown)


dsarrut:
Awesome, thanks !!

uhqd:
Congratulations for fixing this bug !
Hope this will be in the 8.2 release... Thank you very much anyway. 
Small question : there is an internal discussion here to decide where to write the visualisation command line : at its historic place (i.e. at the begining of the .mac) or after the initialize command. Some of us (the religious ones) think that this issue was responsible of this bug. But if I understand well, we can let the visu line at the beginning of the file (please answer, we bet a beer on this issue...). Again thank you.

SimonRit:
No, this issue was not responsible of this bug. The code was testing a boolean which was never set. So it basically relied on what was in memory. Depending on what was done before with the memory now occupied by the boolean, it could have been true (non zero) or false (zero). So you could have different behaviors by switching the order of the commands but that was by luck.
As for the order of the commands, I never know when one is allowed to do what. So I'll let someone else answer that question. Yet, I have always set visu before init.
BTW, I like beer too. Cheers! Simon

uhqd:
Ok thank you. Wait you for a beer here...Cheers

PR #225, opened 2019-02-14
Time Out and Time Production options enabled for the PhaseSpace actor

jawka:
I implemented the "Leaving" time (TOut) and production time (TProd) of the particle according to the paper:
Robert, C. at el. PMB (2013), vol.58(19), pp. 6867-6885.

### DIFF
--- a/source/digits_hits/include/GatePhaseSpaceActor.hh
+++ b/source/digits_hits/include/GatePhaseSpaceActor.hh
@@ -70,6 +70,9 @@ public:
   void SetIsSecStored(bool b){EnableSec = b;}
   void SetIsAllStep(bool b){EnableAllStep = b;}
 
+  void SetIsTOutEnabled(bool b){EnableTOut = b;}
+  void SetIsTProdEnabled(bool b){EnableTProd = b;}
+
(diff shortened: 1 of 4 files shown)


dsarrut:
very good, thanks ! Could you please add few lines in the documentation [here](http://wiki.opengatecollaboration.org/index.php/Users_Guide:Tools_to_Interact_with_the_Simulation_:_Actors#Phase_Space_Actor) to explain this option ? Thanks !

SimonRit:
Thanks. To be honest, I think new items in the PhaseSpace actor should be off by default. Then there is no surprise for existing macros. Is there any reason for having it on by default?

jawka:
@dsarrut I think I don't have an access to the documentation? Should I request for an account or is there any way around to do that?
@SimonRit thanks, for pointing this out. I made them "on" by accident, I didn't realize there is any restriction for that. Do you want me to do a new pull request?

dsarrut:
for the documentation: see http://wiki.opengatecollaboration.org/index.php/Special:RequestAccount

and yes, new option should be off by default, if you dont mind propose a new PR, it will be fine, thanks

PR #226, opened 2019-02-14
Energy spectr

andiresch:
- added option to enable score energy/nucleus in MeV/u
- added unit in one line which was undefined before MeV
- user has to ensure that mass is not zero with filters. Mass ==0 is not checked. Future improvement?


dsarrut:
thanks Andeas !

PR #228, opened 2019-02-19
energySpectrActor: added option to normalize histograms to number of simulated events

andiresch:


source/digits_hits/src/GateEnergySpectrumActorMessenger.cc
```diff
@@ -202,6 +202,12 @@ void GateEnergySpectrumActorMessenger::BuildCommands(G4String base)
   pEnableEnergyPerUnitMassCMD = new G4UIcmdWithABool(bb, this);
   guidance = G4String("Set energy per nucleus");
   pEnableEnergyPerUnitMassCMD->SetGuidance(guidance);
+  
+  
+  bb = base+"/setRelativeToNbPrimaryEvents";
+  pEnableRelativePrimEventsCMD = new G4UIcmdWithABool(bb, this);
+  guidance = G4String("Set values relative to number of primary events");
```
djboersma: I think you can provide a better description. Something like "Normalize all enabled 1D histograms to the number of primary events."

source/digits_hits/src/GateEnergySpectrumActor.cc
```diff
@@ -226,7 +228,14 @@ void GateEnergySpectrumActor::Construct()
 /// Save data
 void GateEnergySpectrumActor::SaveData()
 {
-    
+   if (mEnableRelativePrimEvents){
+    for(std::list<TH1D*>::iterator it=allEnabledTH1DHistograms.begin();it!=allEnabledTH1DHistograms.end();++it)
+      {
+          (*it)->Scale(1./nEvent);
+      }
+       //pEnergySpectrumFluenceTrack->Scale(1./nEvent);
```
djboersma: Could you remove this dead line of code?

source/digits_hits/src/GateEnergySpectrumActorMessenger.cc
andiresch: True. Changed both, description and flag.

source/digits_hits/src/GateEnergySpectrumActor.cc
andiresch: done.

djboersma:
OK, thanks for the fixes!

PR #230, opened 2019-02-20
jobsplitter: add SLURM support

christopheredsall:
This change adds support for the SLRUM job scheduler
(https://slurm.schedmd.com/). The implementation is simply to copy
the OpenPBS pieces and change the name and the directives from
  `#PBS` to `#SBATCH`
and add a new template script


dsarrut:
thanks !

PR #232, opened 2019-03-05
fixed bug: energy was only updated if new track

andiresch:
Energy of particles was only updated when a new track was found. Now it always updates the energy at any step.


djboersma:
Thanks!

PR #250, opened 2019-07-15
Revert "Hackathon verbose"

dsarrut:
Reverts OpenGATE/Gate#249


dsarrut:
revert because lead to bug with qt (dont know why yet)

Issue #211, opened 2019-01-12
Gate needs to use at least the same C++ standard as ROOT


mojca:
MacPorts installs ROOT6 with `-Dcxx17=ON`. As as consequence Gate fails to compile:
```
[...]

PR #252, opened 2019-07-22
Unit tests2

djboersma:
Adding unit tests to Gate, based on "Catch2": https://github.com/catchorg/Catch2

I chose Catch2 because it is header only, I found it simple to use, and it provides all the functionality that I was looking for.
I also looked at the unit testing framework offered by the "Boost" library (https://www.boost.org/doc/libs/1_66_0/libs/test/doc/html/index.html), but when I tried to run my tests on different linux distributions (with different default boost library versions) I ran into compatibility issues, the API has  changed too much in the past 5 years or so, and I didn't see an easy way to hide those incompatibilities with tricks in cmake or other places. With catch2 we don't have any of these troubles.

In order to compile unit tests I needed to reorganize the CMakeLists.txt file. I tried to do this in accordance with lessons learned in the discussion under issue #211 , but it's far from perfect. Suggestions (or commits) towards improvement are very welcome.

This pull request includes the basic setup for the unit tests, plus just a very small number of unit tests. The point is first to agree on the basic setup, then to write many more unit tests. Stuff we should agree on:
* Do you agree that catch2 is good for Gate?
* Are we happy with how the unit tests are included in cmakelists?
* In parallel to unit tests we have the "benchmark tests". They fail with geant 10.5. This highlights the need for improving the benchmark tests (or replace them completely with some more statistically oriented tests).


dsarrut:
Close this attempt. Somehow related to the benchmarks

PR #258, opened 2019-08-28
Noise bug

dsarrut:
When using noise feature in digitizer, when StartTime is not zero, GATE hangs. 

This commit is an attempt to correct it. Warning, not fully validated yet.


dsarrut:
See PR #267

PR #264, opened 2019-10-14
Corrected Generic Repeater documentation

smngr:
Generic repeater documentation required 'time' column adding to documentation, as defined by the function "ReadTimePlacements" in GateMiscFunctions.cc


dsarrut:
excellent, thanks !

PR #266, opened 2019-11-05
GateActorManager: Deactivate previous SD in SetMultiFunctionalDetector

### LINKED ISSUE #257: Memory leak in KillActor
Hi,

the KillActor seems to have a memory leak. I wanted to use the actor to stop the tracking of the back-to-back photons in the detector (cylindricalPET) but at some point in time the simulations on a cluster stop due the high memory demand. In a very simple test using vGATE this seems to increase by 1-2 Mb per second.

Best,
Moritz

cpommranz:
In case a SD is already attached to the logical volume, the SD should be
replaced by a MSD in SetMultiFunctionalDetector. The previously missing
deactivation of the SD could result in a memory leak (see #257).

Since Geant4 doesn't seem to offer a way to remove a SD from the
G4SDManager, we have to deactivate the SD.

Closes #257


dsarrut:
Very good, thanks!

PR #267, opened 2019-11-07
PET debug

dsarrut:
Correction of 2 bugs during PET simulations:
- noise module see #258 
- dead time module. Memory leak issue. 

Both issues are related with construction problem. It seems that memory was not initialized at the right moment (after volume construction). 

Malloc replaced with std vector. 

Will be merged into develop branch once fully validated (Julien/Joey).


djboersma:
This pull request looks a bit scary, because of the very long commit list. But I guess that is because of the long history (first submitted in 2017)? I browsed through the changes in the 13 changed files, and it looks fine to me. Replacing malloc-ed arrays with std::vectors is a very healthy idea. :)

dsarrut:
yes, It scaries me too ;) I should have done shitty things with git, but I dont really know what ...

PR #295, opened 2020-02-19
Add Azure pipeline

tbaudier:
Add Azure pipeline for CI on Mac and Linux.
It compiles with different options (None, Torch, RTK, Optical)
It manages the cache to store root and Geant4

When the pull request will be accepted I will open an account on Azure. For the moment, you can see the results here : https://dev.azure.com/tbaudier/tbaudier/_build/results?buildId=41&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,202 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- develop
+
+
+jobs:
+- job: Build
+  timeoutInMinutes: 360 # how long to run the job before automatically cancelling
+
+  strategy:
+    matrix:
+      Linux:
+        STRATEGY_NAME: 'Linux'
+        imageName: 'ubuntu-latest'
+        GATE_USE_TORCH: 'OFF'
+        GATE_USE_RTK: 'OFF'
+        GATE_USE_OPTICAL: 'OFF'
+      Linux_Torch:
+        STRATEGY_NAME: 'Linux_Torch'
+        imageName: 'ubuntu-latest'
+        GATE_USE_TORCH: 'ON'
+        GATE_USE_RTK: 'OFF'
+        GATE_USE_OPTICAL: 'OFF'
+      Linux_RTK:
+        STRATEGY_NAME: 'Linux_RTK'
+        imageName: 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


SimonRit:
Great! Thanks in particular for the RTK compilation. It would be good if the compilation results were submitted to [cdash](https://my.cdash.org/index.php?project=GATE&date=2020-02-18). That would provide a list of warnings (now) and failing tests (when they'll be in place). See [here](https://github.com/SimonRit/RTK/blob/master/test/ContinuousIntegration/AzurePipelinesBuildAndTest.yml#L120-L122) how it's done on RTK.

dsarrut:
Yes, I registered Thomas on cdash: https://my.cdash.org/index.php?project=GATE

wrzof:
Great news indeed. 

I see that you target  'ubuntu-latest' in your strategy. Can we target a less recent distribution ? It is not easy to install last version of distribution (the older LTS ubuntu maintened seems to be a good target). 

And add centos7 ? This distribution is widely used in HPC system and laboratory (CNRS IN2P3, CERN, ..). For me GATE have to compile on this platform. 

(I make this suggestion but of I do not know the difficulty of the task, sorry if it is too difficult)

tbaudier:
> Great news indeed.
> 
> I see that you target 'ubuntu-latest' in your strategy. Can we target a less recent distribution ? It is not easy to install last version of distribution (the older LTS ubuntu maintened seems to be a good target).
> 
> And add centos7 ? This distribution is widely used in HPC system and laboratory (CNRS IN2P3, CERN, ..). For me GATE have to compile on this platform.
> 
> (I make this suggestion but of I do not know the difficulty of the task, sorry if it is too difficult)

Only Linux Ubuntu is available on Azure: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops
For other Linux distribution, we could run a Docker image on a remote machine: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-linux?view=azure-devops

For Ubuntu I can only choose between ubuntu-18.04 and ubuntu-16.04. I prefer to use latest because I do not know how many time they store these versions of Ubuntu. I just realize that I need to tag the cache with the version of the OS maybe to redo it when the latest OS changed (https://github.com/tbaudier/Gate/blob/develop/azure-pipelines.yml#L111)

dsarrut:
Very good, thanks! Let's start first with the current Ubuntu+OSX. We will look for computing resources for additional linux distributions (centos in particular) with docker a bit later. It should be feasible, but one step at a time ;)

PR #297, opened 2020-02-19
Enable coincidence sorting with generic repeater

cpommranz:
Enable sorting coincidences when genericRepeater is used.

Note that the coincidence sorter does not work for combinations of
angularRepeater and genericRepeater for now.

refs #296


dsarrut:
excellent, thanks !

PR #304, opened 2020-02-24
Remove RPATH for Mac

tbaudier:
With make install on MacOS, Gate could not work. Removing RPATH, it works


dsarrut:
@tbaudier is this PR still valid ?

tbaudier:
Yes, we recently have a similar problem on the Gate mailling list

But I do not know if it's similar or not with https://github.com/OpenGATE/Gate/pull/269

PR #311, opened 2020-02-28
Remove compilation warnings on Linux and Mac

tbaudier:
change int to unsigned int for comparision
define UNUSED function in GateConfiguration.h.in for ununsed variables
remove private ununsed member variable
Change in Torch data to data_ptr


dsarrut:
(RTK will be corrected soon)

PR #313, opened 2020-03-01
Changed if condition to include IAEA phsp

MCvin:
I fixed a condition that was excluding IAEA phase spaces as sources !
But the IAEA output doesn't work either and I couldn't find the problem...


dsarrut:
thanks!

MCvin:
But this only IAEA as source

Le lun. 9 mars 2020 à 09:09, David Sarrut <notifications@github.com> a
écrit :

PR #314, opened 2020-03-07
Update defining_a_system_scanner_ct_pet_spect_optical.rst

MaxTousss:
Explicit in the system_tab table that "scanner" and "PETscanner" are equivalent system except for the creation of a coincidence. Also added a note about level5 component being not available for Gate v8.2.


dsarrut:
thanks!

PR #319, opened 2020-03-09
Add compliance level to be able to compile RTK with ITK

tbaudier:


SimonRit:
FYI, now useless see discussion on SimonRit/RTK#313. You have been too fast!

Issue #320, opened 2020-03-09
cmake complains about "GateLib"


djboersma:
On Ubuntu 18.04 (with cmake version 3.10.2) I cannot build Gate, because cmake runs into this error:

```
CMake Error at CMakeLists.txt:283 (TARGET_LINK_LIBRARIES):
  Object library target "GateLib" may not link to anything.
```

On my personal laptop with a newer OS and cmake 3.13.4 it runs fine, so maybe this error is just a quirk of the cmake version 3.10.2 that happens to be installed on Ubuntu 18.04. But Gate should really be buildable on a vanilla Ubuntu 18.04 system. So either we make it work with 3.10.2 or (less preferred) we explicitly require a "new enough" cmake with the `CMAKE_MINIMUM_REQUIRED` command, using a much more recent version than we do now (3.0).

tbaudier:
I tried to debug but I am not an expert of Cmake.
It seems that Object Librairies (like GateLib) do not have to be linked to anything:
https://stackoverflow.com/questions/53416527/cmake-object-library-target-may-not-link-to-anything

But without the line https://github.com/OpenGATE/Gate/blob/develop/CMakeLists.txt#L283 , it fails to build (see https://github.com/OpenGATE/Gate/pull/318 following https://github.com/OpenGATE/Gate/pull/290)

djboersma:
I really hope that we can make it work for the cmake version that is installed on vanilla Ubuntu 18.04 systems, because that (and even older systems) is what many users use. But simply setting the minimal cmake version to 3.12 seems to be *much* less work for us...
I think that on Ubuntu 18.04, there is an option to install the newest cmake with "snap" instead of with "apt". That might be a solution for some users, but not for all. Some sysadmins are (understandably) quite conservative.

BTW: the `GateLib` object library is now defined in the `ELSE` leg of the `IF(GATE_USE_GPU AND CUDA_FOUND)`, but used outside of that if-else construct. This won't work if anyone tries to build with CUDA. Maybe this is the moment to officially and practically remove the CUDA code, after deprecating its use for more than two years?

tbaudier:
> I really hope that we can make it work for the cmake version that is installed on vanilla Ubuntu 18.04 systems, because that (and even older systems) is what many users use. But simply setting the minimal cmake version to 3.12 seems to be _much_ less work for us...

[...]

PR #322, opened 2020-03-10
Torch yes cuda no

djboersma:
This PR tries to fix issue #320 and the issues that PR #318 tried to address: some struggles between GateLib, cmake versions and libtorch.

In other words: we want to be able to build and link Gate with old cmake (versions older than 3.12) and new cmake (version 3.12 and newer). We also want to be able to build Gate with and without "libtorch".
I tested these 2x2 combinations with this branch and for me it works. Please try on more systems before merging. In order to make a "libtorch" build possible, I needed to add a cmake line to add the include directories. This is necessary if libtorch is not installed in a system directory (i.e. not under `/usr` or `/usr/local`).

Also note a  kind of major tangential issue: the `CMakeLists.txt` file had inconsistent conditional code for building with/without CUDA. Since the CUDA/GPU code has been without maintainer for quite a long time, rather than fixing the inconsistency it seems better to finally remove the GPU code altogether (after a long deprecation period). This affects:
* the references in the `CMakeLists.txt` file
* the `source/gpu` directory (with CUDA kernel code)
* several dozen purely GPU-specific classes
* references in the rest of Gate to the GPU-specific code.

All that is gone now. This hurts a bit, because GPU computing is so tantalizingly cool and it's clear that a lot of effort went into this code. But it's better this way, I think. The code is still available here on GitHub, so if anyone would like to do a new try with CUDA, then maybe some of the old code can serve as learning material. However, the CUDA library today is very different from CUDA back when this code was written. Also, the original authors have stated that they were not happy with the overall design: the Gate/Geant4 environment does not allow proper GPU parallelism, if I understood them correctly. So it's likely that any new attempts to use CUDA in Gate have to be very different.


tbaudier:
The commit https://github.com/OpenGATE/Gate/pull/322/commits/169db37125ef85b4e1e6481f798cd8a8cf62bcac introduces a lot of Warnings from Torch
https://my.cdash.org/buildSummary.php?buildid=1787793

djboersma:
> 
> 
> The commit [169db37](https://github.com/OpenGATE/Gate/commit/169db37125ef85b4e1e6481f798cd8a8cf62bcac) introduces a lot of Warnings from Torch
> https://my.cdash.org/buildSummary.php?buildid=1787793

Yes, I noticed that. I first thought that was more of a problem for torch than for us, because when I looked on my terminal (compiling with gcc 9.2) there were lots of warnings about unused variables. But in your link I see now that there are issues with which C++ standard we use. I wonder if this will be fixed if we simply remove line 300, `set(CMAKE_CXX_STANDARD 11)`. I think that that line was originally introduced to stop users from trying to compile with **too old** compilers (gcc 4.2 and older, I think) which did not support c++11 yet.  The intention was **not** to prohibit c++ standards 14 and 17.
So let me remove that line and see if that reduces the compile warning flood a bit.

djboersma:
Actually, I wonder why all these warnings only popped up with my commit. The only torch-related thing that I did was to make sure that the torch include directory would be included in the `-Iincdir` flags (because without it, it didn't compile for me when I set `GATE_USE_TORCH` to `ON`). The warnings should have been there on your test system before as well, because on your test systems it should still `#include` the same header files, before and after my commit.

Anyway. I think that my commit dams in the flood of warnings a little bit, also by using GCC pragmas to disable certain warnings when including the torch headers. This may not help with other compilers than GCC, such as Clang. And there are still many warnings left, with assert-like macros that are apparently `#define`d with a boolean test and a message implemented through variadic parameters, but in the code often only the boolean expression is given and the message part is left out.

djboersma:
Aha, now the Azure build tests "Linux_Torch" and "MacoS_Torch" have finished and it looks like there are "0 errors / 0 warnings". That is nice. I see that on those systems you use gcc 7.4, while I am using gcc 9.2, which is apparently way more pedantic. I guess it might be useful to mention in the release notes that gcc 7.4 was used as a default compiler, and that with newer compilers such as gcc 9.2 the build may be quite noisy.

djboersma:
Yay! :)

PR #330, opened 2020-04-21
allows to save projection by run

dsarrut:
In NN-ARF, the computed projections filenames now include the run id, if the following flags are set:
/gate/actor/arf/setOverWriteFilesFlag     false
/gate/actor/arf/setResetDataAtEachRunFlag true


dsarrut:
update to come a bit later

PR #331, opened 2020-04-25
Score energy deposition exploiting polar/cylindrical symmetry

andiresch:
This adds GateCylindricalEdepActor, which allows to score the edep/dose along a user defined axis in cylinders. The output is a 2D image with coordinates radius and depth. Currently, only two beam directions are possible: along the x and z axis. As this actor uses existing infrastructure, the output is formally a cartesian coordinate system (3D with one singleton dimension).

Implementation was originally developed and intesnively validated on a branch of Gate 8.1. Only a simple sanity check was performed on this re-implementation in Gate v9.


dsarrut:
thanks !

PR #290, opened 2020-02-13
 Add GateLib library (redo 0cca60010ed825ead69d1fe6b798718c83e149bb)

wrzof:
Avoiding static library, it fix the problem mentionned in https://github.com/OpenGATE/Gate/commit/ef4b8d981a8f2176bed1354ea65262bc6c7c5d93#commitcomment-37267735

PR #336, opened 2020-05-14
Set minimum ITK version to the latest v5.1 and fix compilation with it

SimonRit:
Fix compilation with the new ITK v5.1. The main issues are:
- enum types have been changed in ITK, see InsightSoftwareConsortium/ITK@0dd31a6
- ITKIOGDCM no longer implies ITKGDCM since InsightSoftwareConsortium/ITK@ca1913c
- RTK requires a specific cmake policy for IN_LIST if compiled with CUDA since SimonRit/RTK@0965dfe

Since RTK now requires the latest ITK, I suggest to do the same with Gate, which is similar to the approach used for Geant4, hence the new ITK_MINIMAL_VERSION.
Note that for the compilation I had to uncomment a line in the CMakeLists.txt as discussed in #290 but I did not commit it yet as I don't really understand what's going here...


SimonRit:
This afternoon I corrected my Gate (with a déjà vu feeling) and I was ready to make a nice PR until I found this one... Since no one disagreed, can we merge @dsarrut?

Issue #186, opened 2018-04-29
minimum CMake version


djboersma:
For geant4, the minimum CMake version is 3.3, according to our own install wiki. Shouldn't we then also require that minimum version (or greater) in our own toplevel `CMakeLists.txt` file?

bersbersbers:
Why? What about people who might use Geant4 precompiled binaries and don't compile Geant4 at all? Shouldn't they be able to use as old a Cmake version as possible?

djboersma:
When running cmake for Gate, the G4_INSTALL_PATH/lib64/G4VERSION/*.cmake files need to be read. Those files are written in a build process that relies on a >3.3 cmake install. They do _not_ contain a "CMAKE_MINIMUM_REQUIRED" line, AFAICS in the install that I checked. So it's _conceivable_ that they can be read by older CMake versions. 

I think that the oldest CMake version that I still have on a system accessible by me is 2.8 (linux mint 17.3, based on ubuntu 14.04). I cannot check the build with the 2.4 version of CMake (unless I take the trouble of installing an ancient CMake version just for this purpose, which I am not quite willing to do, my time is limited....).

My worry about allowing older versions is that cmake has "policies" that define how the build is done, and these policies are version dependent. There is a mechanism to deal with that, but this is a lot of extra work. Newer CMake versions also introduce new commands, occasionally, we might want to use those if they are good for something and then it's a hassle to make a workaround for the older cmake versions. We lack manpower, so I'd prefer to limit the allowed cmake versions to a range that we can comfortably support.

dsarrut:
See #337

PR #337, opened 2020-05-14
Fix GateLib dependencies as discussed in #290 #186

SimonRit:


djboersma:
When I try this on a Linux Mint 19 system (similar to Ubuntu 18.04) with cmake version 3.10.2 then I get this error:
```
CMake Error at CMakeLists.txt:265 (TARGET_LINK_LIBRARIES):
  Object library target "GateLib" may not link to anything.
```

djboersma:
On a system with cmake version 3.13.4 it seems to work fine.

djboersma:
Strange. One of the 9 build bots is an Ubuntu Linux system with cmake version 3.10.2 as well. It has no problem building this PR. So now I'm confused. @tbaudier , is it indeed that cmake version on that build bot, or are you installing a newer cmake somehow?

tbaudier:
Target_Link_Libraries for object libraries was introduced with cmake version 3.12:
https://cmake.org/cmake/help/v3.12/release/3.12.html#commands

For Azure, I did the PR https://github.com/OpenGATE/Gate/pull/338 to display the cmake version. You right, Ubuntu installed version 3.10 (https://dev.azure.com/OpenGateCollaboration/Gate/_build/results?buildId=105&view=logs&j=2f953adc-c56d-55c4-a64a-eab7c4b02235&t=b7cd74e9-98b2-5d11-874b-a97b8f10da9a&l=290) via package manager but this is the version 3.17 used to compile (https://dev.azure.com/OpenGateCollaboration/Gate/_build/results?buildId=105&view=logs&j=2f953adc-c56d-55c4-a64a-eab7c4b02235&t=b7cd74e9-98b2-5d11-874b-a97b8f10da9a&l=570). I do not know why ...

djboersma:
But this means that we can only allow this PR if at the same time we raise the "cmake minimum version" to 3.12. Normally we try to be nice to users who work with slightly older platforms. Alternatively, we could keep both the old and the new linking code, and selecting either of them with an `if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")` condition. But I'd not be in favor of that.

Maybe @dsarrut can guide us here?

SimonRit:
Thanks for all the testing. I think cmake is very easy to update on any platform. Another option is to rollback to a dynamic library but one would need to fix the problem @dsarrut mentioned [here](https://github.com/OpenGATE/Gate/commit/ef4b8d981a8f2176bed1354ea65262bc6c7c5d93#commitcomment-37267735).

SimonRit:
Maybe one more comment from an outsider: I am not convinced that the new binaries make sense. It is not clear from their name but they seem to be processing tools for a specific application, the Compton camera. Currently, with the use of an object library, everybody gets 5 exe instead of 1, all as large as Gate. But I guess this choice has been discussed within the steering committee...

tbaudier:
> Maybe one more comment from an outsider: I am not convinced that the new binaries make sense. It is not clear from their name but they seem to be processing tools for a specific application, the Compton camera. Currently, with the use of an object library, everybody gets 5 exe instead of 1, all as large as Gate. But I guess this choice has been discussed within the steering committee...

I did this PR https://github.com/OpenGATE/Gate/pull/343 to compile the Compton Camera binaries on demand in CMake

djboersma:
The five binaries for a special application may seem a bit excessive, but I am actually quite happy with splitting the monolithic Gate into a main executable and a library. If we can agree on a way forward with `CMake`, then I can finally start merging the unit test framework into the develop branch, because that also relies a lot on being able to link against `GateLib`.

If disk space is deemed an issue, then we should maybe look at dynamic libraries instead of static ones. ROOT and Geant4 also organize their functionality in several `*.so` files. Especially the ROOT guys care a lot about speed, so probably the performance hit from dynamic linking is either negligible or avoidable. Of course we may be opening a whole new time consuming can of worms, so I'm not sure how realistic this option is, manpower-wise. Still it would be nice if some pioneer with a few hours/days to spare could give it a try this summer. We might be positively surprised.

tbaudier:
I rebased the PR branch with the current develop branch to remove conflicts.
I also added a commit to update the minimal cmake version

dsarrut:
thanks !

PR #339, opened 2020-05-14
Changed all ROOT SetAutoSave to 1e9 to reduce disk writing

MCvin:


dsarrut:
Thanks Maxime! We will test it ... On your side, did you manage some changes?

JoeyLabour:
Thank you. I already launched some tests for PET simulations.

dsarrut:
Tests show no change in simulation speed, so we leave it like it was. Thanks all for these experiments !

Issue #341, opened 2020-06-03
Phase Space Actor with multiple time slices


fkhellaf:
Hi,

I encountered an issue using Gate v9.0. There is a crash when I use a Phase Space Actor and more than one time slice. The problem does not occur when there is only one time slice, or if I remove the actor and use several slices.
Here is a minimal example:

# Geometry
/gate/geometry/setMaterialDatabase  data/GateMaterials.db
/gate/world/setMaterial             Vacuum
/gate/world/geometry/setXLength     20 m
/gate/world/geometry/setYLength     20 m
/gate/world/geometry/setZLength     20 m

# Physics list
/gate/physics/addPhysicsList QGSP_BIC

# Phase spaces
/control/alias DetectorName PhaseSpaceIn

/gate/world/daughters/name                            Plane{DetectorName}
/gate/world/daughters/insert                          box
/gate/Plane{DetectorName}/geometry/setXLength         1 nm
/gate/Plane{DetectorName}/geometry/setYLength         1000 mm
/gate/Plane{DetectorName}/geometry/setZLength         1000 mm
/gate/Plane{DetectorName}/setMaterial                 Vacuum
/gate/Plane{DetectorName}/vis/setVisible              1
/gate/Plane{DetectorName}/vis/setColor                yellow

/gate/actor/addActor PhaseSpaceActor                  {DetectorName}
/gate/actor/{DetectorName}/save                       output/{DetectorName}.root
/gate/actor/{DetectorName}/attachTo                   Plane{DetectorName}
/gate/actor/{DetectorName}/enableEkine                true
/gate/actor/{DetectorName}/enableXPosition            false
/gate/actor/{DetectorName}/enableYPosition            true
/gate/actor/{DetectorName}/enableZPosition            true
/gate/actor/{DetectorName}/enableXDirection           true
/gate/actor/{DetectorName}/enableYDirection           true
/gate/actor/{DetectorName}/enableZDirection           true
/gate/actor/{DetectorName}/enableProductionVolume     false 
/gate/actor/{DetectorName}/enableProductionProcess    false
/gate/actor/{DetectorName}/enableParticleName         false
/gate/actor/{DetectorName}/enableWeight               false
/gate/actor/{DetectorName}/enableTime                 true
/gate/actor/{DetectorName}/storeSecondaries           true
/gate/actor/{DetectorName}/useVolumeFrame             false
/gate/actor/{DetectorName}/storeOutgoingParticles     false
[...]

PR #342, opened 2020-06-08
Debug PhaseSpace writting for multiple slice time

tbaudier:
With multiple time slice, the mFile in GatePhaseSpaceActor was close at the end of one time slice instead of write.
So I close the mFile in the destructor and create a write function in GateTreeFile classes

Debug https://github.com/OpenGATE/Gate/issues/341

@fkhellaf , can you run simulation to be sure it works?
@wrzof , are you agree with that PR?

Thank you

### DIFF
--- a/source/digits_hits/src/GatePhaseSpaceActor.cc
+++ b/source/digits_hits/src/GatePhaseSpaceActor.cc
@@ -111,6 +111,7 @@ GatePhaseSpaceActor::~GatePhaseSpaceActor()
   free(pIAEARecordType);
   pIAEAheader = 0;
   pIAEARecordType = 0;
+  mFile.close();
   delete pMessenger;
   GateDebugMessageDec("Actor", 4, "~GatePhaseSpaceActor() -- end\n");
 }
@@ -693,7 +694,7 @@ void GatePhaseSpaceActor::SaveData()
     fclose(pIAEARecordType->p_file);
   }
   else {
-    mFile.close();
+    mFile.write();
   }
 }
 // --------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


wrzof:
Hi, 

Why adding a write method ? For me, move the call to `close` method from `SaveData` to the destructor of `GatePhaseSpaceActor` will be sufficient no ? 

Specially because most of `write` functions are empty and I think `Close` in ROOT files make a call to `Write`.

tbaudier:
Hi,

Good idea, but I think there is a reason (I ignored it) the function ```GatePhaseSpaceActor::SaveData()```exits.
Before your commit https://github.com/OpenGATE/Gate/commit/43cc7c7933ea915a348c3cef00727deaa28a84ed  , the root file was saved and not closed in that function https://github.com/OpenGATE/Gate/commit/43cc7c7933ea915a348c3cef00727deaa28a84ed#diff-756c17058b97baf12d072873429c8ca4L607
I think there is a reason why it was commented here: https://github.com/OpenGATE/Gate/commit/1dede7b0732ce7b584502d6a81ffb2890340b734#diff-756c17058b97baf12d072873429c8ca4L316

Thomas

fkhellaf:
Hi,

The simulation works fine now, thanks!

Feriel

wrzof:
> Good idea, but I think there is a reason (I ignored it) the function `GatePhaseSpaceActor::SaveData() `exists.

I don't know if you still ignore it, but `SaveData` is useful if you need to save one image (or whatever) per run (or event) or something like that because this function is called in `EndOfRunAction` and `EndOfEventAction`. Which is not the case here. `

Or maybe you want to force the flush of file times to times (I guess that what `TFile::Write` does) . I don't know if it makes sense here, it can create unnecessary write to disk. 

> Before your commit 43cc7c7 , the root file was saved and not closed in that function 43cc7c7#diff-756c17058b97baf12d072873429c8ca4L607

You are right, I did it wrong. 

So for me, in this PR, we only need to keep the move of `close` in the destructor.

dsarrut:
thanks! (save must be kept in SaveData, the user may want to save per run sometimes)

Issue #335, opened 2020-05-08
empty ascii (and tree) hits output when using PhaseSpaceActor


mcristof:
Hi all,

when I add a PhaseSpaceActor 

/gate/actor/addActor PhaseSpaceActor                PhaseSpace1
/gate/actor/PhaseSpace1/attachTo                    phantom
/gate/actor/PhaseSpace1/save                        output/phs.root

the output of
/gate/output/ascii/enable
/gate/output/ascii/setFileName   output/test
/gate/output/ascii/setOutFileHitsFlag   1

(and the same for /gate/output/tree/)
is an empty file

without that Actor everything is ok.

Here the full code:

#=====================================================
# GEOMETRY
#=====================================================
/gate/geometry/setMaterialDatabase data/GateMaterials.db
/gate/world/geometry/setXLength 1. m
/gate/world/geometry/setYLength 1. m
/gate/world/geometry/setZLength 1. m
/gate/world/setMaterial Air
/gate/world/daughters/name              scanner
/gate/world/daughters/insert            box
/gate/scanner/geometry/setXLength       1 m
/gate/scanner/geometry/setYLength       1 m
/gate/scanner/geometry/setZLength       1 m
/gate/scanner/placement/setTranslation  0.0 0.0 0.0 cm
/gate/scanner/setMaterial               Air
/gate/scanner/daughters/name            phantom
/gate/scanner/daughters/insert          box
/gate/phantom/geometry/setXLength       40 cm
/gate/phantom/geometry/setYLength       40 cm
/gate/phantom/geometry/setZLength       1 cm
/gate/phantom/placement/setTranslation  0.0 0.0 10.0 cm
/gate/phantom/setMaterial               Water
/gate/phantom/repeaters/insert          linear
/gate/phantom/linear/setRepeatNumber    20
/gate/phantom/linear/setRepeatVector    0. 0. 10. mm
/gate/systems/scanner/level1/attach     phantom
/gate/phantom/attachCrystalSD
/gate/physics/addPhysicsList emstandard
#=====================================================
# ACTORS
#=====================================================
/gate/actor/addActor               SimulationStatisticActor stat
/gate/actor/stat/save              output/stat-proton.txt
/gate/actor/addActor PhaseSpaceActor                PhaseSpace1
/gate/actor/PhaseSpace1/attachTo                    phantom
/gate/actor/PhaseSpace1/save                        output/phs.root
/gate/actor/PhaseSpace1/enableProductionProcess     true
[...]

PR #354, opened 2020-09-08
GateActorManager: Replace previous SD in SetMultiFunctionalDetector

cpommranz:
Replace the currently attached SD with a cloned and inactive one. The original and active SD is later used in the multi sensitive detector.

For more detailed information see the commit messages.

This is a rather dirty hack relying on implicit Geant4 behavior. If anyone comes up with a better solution, I will very much appreciate it.

Refs #350, #335

### DIFF
--- a/source/arf/include/GateARFSD.hh
+++ b/source/arf/include/GateARFSD.hh
@@ -70,6 +70,8 @@ public:
   //! Destructor
   ~GateARFSD();
 
+  GateARFSD* Clone() const override;
+
   //! Method overloading the virtual method Initialize() of G4VSensitiveDetector
   void Initialize(G4HCofThisEvent*HCE);
 
--- a/source/arf/src/GateARFSD.cc
+++ b/source/arf/src/GateARFSD.cc
@@ -79,6 +79,46 @@ GateARFSD::~GateARFSD()
   delete mArfTableMgr;
   }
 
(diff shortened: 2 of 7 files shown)


dsarrut:
A huge thanks to Christian for this PR. This is a tricky issue. I don't have a better solution right now. We keep in mind to ask Geant4 people to expose the function (RemoveSD).

dsarrut:
Additional comment for record: 
Susanna: "Makoto said that the target object of the G4SDStructure is not publicly visible, therefore to make the RemoveSD(0) method public does not help. What should be done is to set a new cloned sensitive detector to G4SDManager. Then, the old SD is automatically removed. This is in the assumption that the cloned SD has the same name as the original SD."

Christian: "Adding a cloned sensitive detector to the G4SDManager is exactly what I did in the Gate PR. It's good to know that this is the recommended way. I was a bit unsure because of the semantic incorrectness (adding a cloned but deactivated sensitive detector vs. "removing"). It's correct that making the G4SDStructure's RemoveSD method public does not directly help. Sorry, I was unprecise in the PR message. I was thinking of adding a "RemoveDetector" (or similar) method to G4SDManager in a similar way to "G4SDManager::AddNewDetector(G4VSensitiveDetector*aS)", which would directly or indirectly call the G4SDStructure::RemoveSD method. ... "

PR #363, opened 2020-10-12
Change RadioactiveDecay (Deprecated) to Radioactivation

BishopWolf:
G4RadioactiveDecay is deprecated and will be removed in next version of Geant4

It is recommended to use G4Radioactivation to replace the deprecated code while keeping the old behaviour.


dsarrut:
thanks !

PR #370, opened 2020-11-03
Implementation of three physics lists

dhaberl:
Hello, 

I added three electromagnetic physics constructors which were not present:
1) Single scattering = emstandard_SS
2) Goudsmit Saunderson = emstandard_GS
3) Low-energy = emLowEP

I tested it on my machine and it works fine.

Best,
David


dsarrut:
very good, thanks! I think single scattering list was asked by @GitFuchs also ...

tbaudier:
This commit added .DS_Store binariy files, is it possible to remove them?

PR #371, opened 2020-11-09
Update digitizer_and_detector_modeling.rst

CarstenHinz:
Changed inverse square law to implementation within code.
see: https://github.com/OpenGATE/Gate/blob/develop/source/digits_hits/src/GateInverseSquareBlurringLaw.cc


dsarrut:
thank you!

PR #372, opened 2020-11-11
New output module and other modifications

Crauzerus:
1. **New output module - GateExtendedAnalysis** 

GateAnalysis class is not compatible with GateExtendedVSource class output because does not provide a proper statistics about compton and rayleigh scatterings - it focuses only on the 2 gammas aspect.

GateExtendedAnalysis class solves this problem and provides these statistics for all gammas emitted by GateExtendedVSource class.

2. **New branches**

To generated output files by GateToTree and GateToRoot modules are added three new branches: sourceKind, decayModel and gammaKind (all represent an additional information about gamma provided by GateExtendedVSource class).

3. **New commands for GateToTree class**

GateToTree provides only an option to disable a chosen branch in the tree. In many cases this means writing a lot of lines of code in the macro to reduce a tree's size. Due to that fact, I added two new kinds of commands:

- _disable all branches_ - with this command user can disable all branches from a tree,
- _enable a branch_ - with this command user can enable a chosen branch

GateToTree branches names for volumes (e.g. volume[1]) and systems (e.g. volumeName/level1) are not compatible with ROOT. 

Format with square brackets is impossible to read with ROOT C++ and pyROOT.

Format with a slash causes warnings during GATE works and errors during work with ROOT.

To solve this problem I added a command _enableRootFriendlyFormat_ which changes these formats to "_" (e.g. volumeID_1, volumeName_level1"). By default this option is disable.

4. **Other modifications**:
- fixed memory leaks in the GateToTreeMessenger class
- added an initialization of variables in the GateCrystalHit class - not initialized ones cause a random values
- replaced new/delete with unique_ptr


dsarrut:
Thanks a lot, Mateusz! Could you please split the PR into 3 (or 4), according to the points you mentioned? Also, can you provide tests in GateContrib for all those 3 PRs ? 
Maybe a detail: rename "Kind" into "Type" for sourceKind, gammaKind ...? Not clear what it represents for me.

Crauzerus:
> Thanks a lot, Mateusz! Could you please split the PR into 3 (or 4), according to the points you mentioned? Also, can you provide tests in GateContrib for all those 3 PRs ?
> Maybe a detail: rename "Kind" into "Type" for sourceKind, gammaKind ...? Not clear what it represents for me.

Ok, no problem. So I close this PR and do this as you said.

PR #373, opened 2020-11-12
Remove binary files .DS_Store

tbaudier:
Added with PR #370
I just remove the files without rebase the repository to avoid to break the history


SimonRit:
LGTM but we also suggested to add the file name to .gitignore?

SimonRit:
Looks good to me

PR #374, opened 2020-11-12
Try to debug azure cache

tbaudier:
**DO NOT MERGE THIS PR NOW**

Simulate a PR to debug azure failure with cache

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -191,6 +191,20 @@ jobs:
     displayName: 'Install gate compilation dependencies'
 
   - script: |
+      echo "-------------"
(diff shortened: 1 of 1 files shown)


tbaudier:
The runs on Azure pipelines failed because Gate did not find Geant4 and Root folders. Geant4 and Root folders were downloaded from the cache but it seems that during the cache recovery, that folders were not download in the correct path. 
This issue was reported by https://github.com/microsoft/azure-pipelines-tasks/issues/13879

I can do the modification to be able to Gate to compile again but Azure could modify the cavhe behavior again to debug this problem.

PR #391, opened 2020-12-18
merge, update, correct the materials database file

djboersma:
**PLEASE DO NOT YET MERGE, FIRST LOOK AT THE ISSUES/QUESTIONS BELOW**

I ran `md5sum $(locate /home/boersma/*GateMaterials.db) | cut -f1 -d' ' | sort | uniq`
on my number cruncher and discovered that I had 26 different versions of
"the GateMaterials.db file".  Some are in my Gate source trees (of various Gate
versions), but mostly they come from the GateContrib tree.  Most differences between
databases are the insignificant differences in the amount of white space. But
there are also many differences in which elements and materials are included,
and the numerical values of their properties.

I have tried to merge (almost) all databases that I could find.  If the
issues/questions below can be resolved, and if enough experts confirm to
me that they are OK with this list of materials, then I would like to go
ahead and replace the GateMaterials.db files in GateContrib with this
file.

Some observations/comments from the merging process:

* Many database files still contain the old wrong LYSO definition
* I found two different definitions of solid copper, one with 8.96 and one with 8.92 g/cm3.
  I chose the first one, because it seems consistent with what geant4
  uses for G4_Cu.
* Many/most GateMaterials.db copies have "Vanadium" misspelled as "Vandium",
  I am using the correct spelling now.
* For the solid state I found different spellings: "solid", "solide", "Solid" and "Solide".
  I replaced them all with "solid".
* Silver (element) was reported both with 107.868 an 107.87 g/mole, using 107.868 now (NIST).
* Tin (element) was reported both with 118.69 an 118.71 g/mole, using 118.71 now (NIST)
* Whoever added "Antimoine" to their database used the wrong Z, namely 52 (which is actually Tellurium)!!!
  It should be 51. Plus I replaced the French name with the English name ("Antimony").
* I found a db with *lots* of Gold compounds that looked very special,
  application specific. So I left those out.
  (Turns out this was from a db that someone sent to the mailing list,
   as part of a "dose actor problem". Not part of GateContrib.)
* I tried to keep a consistent whitespace policy
  - no tabs
  - no spaces at the end of any line
  - one empty line between subsequent materials
  - for the elements in a mixture compound:
    the "+" of "+el" on position 9,
    the ";" before the "f=" or "n=" on position 30.

I ran into a couple of issues and questions that should be answered
before the merged database should be accepted as "the Gate material
database". Please answer/comment:

* I found three three slightly different versions of PMMA:
  two are quite similar, with slightly different densities and material fractions,
  the third one has Sulphur instead of Oxygen.
* I found three different versions of "air":
  two with similar but different normal densities (1.2-1.3 mg/cm3),
  and one with "double" density (pressurized air?).
  Density actually fluctuates quite a bit with varying ambient temperature and pressure,
  maybe it is normal/useful to have a range of air-like materials for this?
* I found two different versions of NaI(Tl) crystals, with different densities (3.76 or 3.67 g/cm3)
  and different elemental mass fractions.
  The density value 3.67 agrees best with pure sodium iodide (and with densities quoted online by vendors).
  Maybe there really are different types of these crystals?
  Or is one of them just wrong?
* We have now TiO2 (defined as a mixture, with weight fractions) and TiO (defined as a compound, with number of elements), do we really need both? It *is* a compound, right?
* There are now two very different "adipose" tissue materials (differ about a factor of 2 in density), maybe one is just wrong?
* I added/merged most new materials that I found, but I am not quite
  sure if we should really keep all of them. Please have a look and if
  some materials look too exotic or "one time use" then please let me
  know.
* I found a very special GateMaterials.db file in GateContrib, which has only the elements in
  common with the other GateMaterials.db files, and of the "materials" only Water and Air.
  Instead it has definitions of very many human tissues. Should we include all that?
  See GateContrib/dosimetry/doseactor/voxelized-phantom/data/GateMaterials.db

Less urgent questions (at least for the short term):

* Should we update the element molar weights with the more exact NIST/Geant4 values?
  For quite a few elements/materials there seem to be some extra digits of precision available.
* Many materials, including basic materials like water and air, also exist in the G4 database.
  If this is "historic", should we "respect" that or should we "clean up"?
  I am tempted to encourage Gate users to use standard G4 materials whenever possible,
  and only when G4 does not provide what you need, define something new.
* Biological tissues have often a couple of extra elements with f=0.0, what's up with that? Shall we clean that up?
* There is quite a zoo of tissue definitions.
  Sometimes unclear to which are of general use, and which should be regarded as one-time trials.
  How selective/permissive should we be?
* Should we sort the materials list thematically?
  (E.g. elemental materials, detector materials, construction/shielding materials, medical tissues, etc.)

### DIFF
--- a/GateMaterials.db
+++ b/GateMaterials.db
@@ -24,7 +24,7 @@ Potassium:  S= K   ; Z= 19. ; A=  39.098 g/mole
 Calcium:    S= Ca  ; Z= 20. ; A=  40.08  g/mole
 Scandium:   S= Sc  ; Z= 21. ; A=  44.956 g/mole
 Titanium:   S= Ti  ; Z= 22. ; A=  47.867 g/mole
-Vandium:    S= V   ; Z= 23. ; A=  50.942 g/mole
+Vanadium:   S= V   ; Z= 23. ; A=  50.942 g/mole
 Chromium:   S= Cr  ; Z= 24. ; A=  51.996 g/mole
 Manganese:  S= Mn  ; Z= 25. ; A=  54.938 g/mole
 Iron:       S= Fe  ; Z= 26. ; A=  55.845 g/mole
@@ -493,7 +493,7 @@ Epidermis: d=0.92 g/cm3 ; n=11
         +el: name=Calcium    ; f=0.001
         +el: name=Scandium   ; f=0.0
         +el: name=Titanium   ; f=0.0
-        +el: name=Vandium    ; f=0.0
+        +el: name=Vanadium   ; f=0.0
         +el: name=Chromium   ; f=0.0
         +el: name=Manganese  ; f=0.0
 
@@ -506,7 +506,7 @@ Hypodermis: d=0.92 g/cm3 ; n=11
         +el: name=Calcium    ; f=0.001
         +el: name=Scandium   ; f=0.0
         +el: name=Titanium   ; f=0.0
-        +el: name=Vandium    ; f=0.0
+        +el: name=Vanadium   ; f=0.0
         +el: name=Chromium   ; f=0.0
         +el: name=Manganese  ; f=0.0
 
@@ -790,7 +790,7 @@ Tumor: d=1 g/cm3 ; n=11
         +el: name=Calcium    ; f=0.001
         +el: name=Scandium   ; f=0.0
         +el: name=Titanium   ; f=0.0
-        +el: name=Vandium    ; f=0.0
+        +el: name=Vanadium   ; f=0.0
         +el: name=Chromium   ; f=0.0
         +el: name=Manganese  ; f=0.0
 
--- a/GateMaterials.db
+++ b/GateMaterials.db
@@ -135,15 +135,10 @@ NaI: d=3.67 g/cm3; n=2;  state=solid
         +el: name=Sodium     ; n=1
         +el: name=Iodine     ; n=1
 
-NaITlA: d=3.67 g/cm3; n=3;  state=solid
+NaITl: d=3.67 g/cm3; n=3;  state=solid
         +el: name=Sodium     ; f=0.152
         +el: name=Iodine     ; f=0.838
         +el: name=Thallium   ; f=0.010
-
-NaITlB: d=3.76 g/cm3; n=3; state=solid
-        +el: name=Sodium     ; f=0.1531
-        +el: name=Iodine     ; f=0.84547
-        +el: name=Thallium   ; f=0.00136
 
 CsI: d=4.51 g/cm3; n=2;  state=solid
         +el: name=Cesium     ; n=1
@@ -235,11 +230,6 @@ Nomex: d=0.95 g/cm3; n=4; state=solid
 NitrogenGas: d=1.29 mg/cm3 ; n=1 ; state=gas
         +el: name=Nitrogen   ; f=1
 
-Hostapan: d=1.4 g/cm3; n=3; state=solid
-        +el: name=Hydrogen   ; f=0.041959
-        +el: name=Carbon     ; f=0.625017
-        +el: name=Oxygen     ; f=0.333025
-
 Cerrotru: d=8.72 g/cm3; n=2 ; state=solid
         +el: name=Bismuth    ; f=0.58
         +el: name=Tin        ; f=0.42
@@ -253,24 +243,6 @@ Breast: d=1.020 g/cm3 ; n = 8
         +el: name=Sodium     ; f=0.0010
         +el: name=Phosphor   ; f=0.0010
         +el: name=Chlorine   ; f=0.0010
-
-Air: d=1.29 mg/cm3 ; n=4 ; state=gas
-        +el: name=Nitrogen   ; f=0.755268
-        +el: name=Oxygen     ; f=0.231781
-        +el: name=Argon      ; f=0.012827
-        +el: name=Carbon     ; f=0.000124
-
-Air2: d=2.58 mg/cm3 ; n=4 ; state=gas
-        +el: name=Nitrogen   ; f=0.755268
-        +el: name=Oxygen     ; f=0.231781
-        +el: name=Argon      ; f=0.012827
-        +el: name=Carbon     ; f=0.000124
-
-Air3: d=1.20479 mg/cm3 ; n=4 ; state=gas
-        +el: name=Nitrogen   ; f=0.755268
-        +el: name=Oxygen     ; f=0.231781
-        +el: name=Argon      ; f=0.012827
-        +el: name=Carbon     ; f=0.000124
 
 Glass: d=2.5 g/cm3; n=4;  state=solid
         +el: name=Sodium     ; f=0.1020
@@ -372,12 +344,6 @@ LYSO:   d=7.36 g/cm3; n=4 ; state=solid
         +el: name=Silicon    ; f=0.063714272
         +el: name=Oxygen     ; f=0.181479788
 
- LYSOold:   d=5.37 g/cm3; n=4 ; state=solid
-        +el: name=Lutetium   ; f=0.31101534
-        +el: name=Yttrium    ; f=0.368765605
-        +el: name=Silicon    ; f=0.083209699
-        +el: name=Oxygen     ; f=0.237009356
-
 LYSO-Ce-Hilger:   d=7.10 g/cm3; n=5 ; state=solid
         +el: name=Lutetium   ; f=0.713838658203075
         +el: name=Yttrium    ; f=0.040302477781781
@@ -385,11 +351,7 @@ LYSO-Ce-Hilger:   d=7.10 g/cm3; n=5 ; state=solid
         +el: name=Oxygen     ; f=0.181501252152072
         +el: name=Cerium     ; f=0.000635804578835201
 
-LaBr3: d=5.29 g/cm3 ; n=2
-        +el: name=Bromine    ; n=3
-        +el: name=Lanthanum  ; n=1
-
-LaBr3_VLC: d=5.06 g/cm3 ; n=2
+LaBr3: d=5.06 g/cm3 ; n=2
         +el: name=Bromine    ; n=3
         +el: name=Lanthanum  ; n=1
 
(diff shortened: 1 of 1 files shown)


GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
+        +el: name=Oxygen     ; f=0.1538
+        +el: name=Chlorine   ; f=0.0681
+
+Gypsum: d=2.32 g/cm3 ; n=4 ; state=solid
+        +el: name=Hydrogen   ; f=0.023416
+        +el: name=Oxygen     ; f=0.557572
+        +el: name=Sulfur     ; f=0.186215
+        +el: name=Calcium    ; f=0.232797
+
+Beton: d=2.35 g/cm3 ; n=10 ; state=solid
```
SimonRit, line 837: In english, Beton is Concrete

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
+        +el: name=Sodium     ; f=0.0162
+        +el: name=Magnesium  ; f=0.0024
+        +el: name=Aluminium  ; f=0.0433
+        +el: name=Silicon    ; f=0.3008
+        +el: name=Sulfur     ; f=0.0013
+        +el: name=Potassium  ; f=0.0182
+        +el: name=Calcium    ; f=0.0787
+        +el: name=Iron       ; f=0.0118
+
+Beton2: d=3.50 g/cm3 ; n=10 ; state=solid
```
SimonRit, line 849: Beton2 -> Concrete2

GateMaterials.db
```diff
@@ -88,7 +135,31 @@ NaI: d=3.67 g/cm3; n=2;  state=solid
         +el: name=Sodium     ; n=1
         +el: name=Iodine     ; n=1
 
-PWO: d=8.28 g/cm3; n=3 ; state=Solid
+NaITlA: d=3.67 g/cm3; n=3;  state=solid
+        +el: name=Sodium     ; f=0.152
+        +el: name=Iodine     ; f=0.838
+        +el: name=Thallium   ; f=0.010
+
+NaITlB: d=3.76 g/cm3; n=3; state=solid
```
SimonRit, line 143: Maybe chose A or B?

GateMaterials.db
```diff
@@ -108,29 +179,71 @@ Plexiglass: d=1.19 g/cm3; n=3; state=solid
...
+Nomex: d=0.95 g/cm3; n=4; state=solid
+        +el: name=Hydrogen   ; n=10
+        +el: name=Carbon     ; n=14
+        +el: name=Nitrogen   ; n=2
+        +el: name=Oxygen     ; n=2
+
+NitrogenGas: d=1.29 mg/cm3 ; n=1 ; state=gas
+        +el: name=Nitrogen   ; f=1
+
+Hostapan: d=1.4 g/cm3; n=3; state=solid
```
SimonRit, line 238: Any clue what that is?

GateMaterials.db
```diff
@@ -147,25 +260,45 @@ Air: d=1.29 mg/cm3 ; n=4 ; state=gas
         +el: name=Argon      ; f=0.012827
         +el: name=Carbon     ; f=0.000124
 
+Air2: d=2.58 mg/cm3 ; n=4 ; state=gas
```
SimonRit, line 263: Favor G4_AIR

GateMaterials.db
```diff
@@ -147,25 +260,45 @@ Air: d=1.29 mg/cm3 ; n=4 ; state=gas
...
+        +el: name=Nitrogen   ; f=0.755268
+        +el: name=Oxygen     ; f=0.231781
+        +el: name=Argon      ; f=0.012827
+        +el: name=Carbon     ; f=0.000124
+
+Air3: d=1.20479 mg/cm3 ; n=4 ; state=gas
```
SimonRit, line 269: Favor G4_AIR

GateMaterials.db
```diff
@@ -220,22 +366,55 @@ PTFE:   d= 2.18 g/cm3 ; n=2 ; state=solid
...
-        +el: name=Silicon; f=0.063714272
-        +el: name=Oxygen; f=0.181479788
+LYSO:   d=7.36 g/cm3; n=4 ; state=solid
+        +el: name=Lutetium   ; f=0.714467891
+        +el: name=Yttrium    ; f=0.04033805
+        +el: name=Silicon    ; f=0.063714272
+        +el: name=Oxygen     ; f=0.181479788
 
- LYSOold:   d=5.37 g/cm3; n=4 ; state=Solid
+ LYSOold:   d=5.37 g/cm3; n=4 ; state=solid
```
SimonRit, line 375: I would remove

GateMaterials.db
```diff
@@ -220,22 +366,55 @@ PTFE:   d= 2.18 g/cm3 ; n=2 ; state=solid
...
+        +el: name=Yttrium    ; f=0.040302477781781
+        +el: name=Silicon    ; f=0.063721807284236
+        +el: name=Oxygen     ; f=0.181501252152072
+        +el: name=Cerium     ; f=0.000635804578835201
+
+LaBr3: d=5.29 g/cm3 ; n=2
+        +el: name=Bromine    ; n=3
+        +el: name=Lanthanum  ; n=1
+
+LaBr3_VLC: d=5.06 g/cm3 ; n=2
```
SimonRit, line 392: Any clue what's VLC?

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
         +el: name=Chlorine   ; f=0.002000
         +el: name=Potassium  ; f=0.002000
 
-PMMA:   d=1.195 g/cm3; n=3 ; state=Solid
+SoftTissue: d=1.04 g/cm3 ; n=16
```
SimonRit, line 667: Prefer G4_TISSUE_SOFT_ICRP?

GateMaterials.db
```diff
@@ -292,10 +471,19 @@ Adipose: d=0.92 g/cm3 ; n=11
         +el: name=Calcium    ; f=0.001
         +el: name=Scandium   ; f=0.0
         +el: name=Titanium   ; f=0.0
-        +el: name=Vandium    ; f=0.0
+        +el: name=Vanadium   ; f=0.0
         +el: name=Chromium   ; f=0.0
         +el: name=Manganese  ; f=0.0
 
+Adipose_Tissue: d=1.95 g/cm3 ; n=7
```
SimonRit, line 478: Prefer G4_ADIPOSE_TISSUE_ICRP?

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
+        +el: name=Calcium    ; f=0.0235
+
+LungEquiv: d=0.30 g/cm3 ; n=7
+        +el: name=Hydrogen   ; f=0.0836
+        +el: name=Carbon     ; f=0.6041
+        +el: name=Nitrogen   ; f=0.0167
+        +el: name=Oxygen     ; f=0.1733
+        +el: name=Magnesium  ; f=0.1136
+        +el: name=Silicon    ; f=0.0072
+        +el: name=Chlorine   ; f=0.0015
```
SimonRit, line 749: I don't understand what are those materials

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
-Epoxy:  d=1.0 g/cm3; n=3; state=Solid
+PMMA2:   d=1.19 g/cm3 ; n=3 ; state=solid
+        +el: name=Hydrogen   ; f=0.080538
+        +el: name=Carbon     ; f=0.599848
+        +el: name=Oxygen     ; f=0.319614
+
+PMMA3:   d=1.195 g/cm3; n=3 ; state=solid
+        +el: name=Hydrogen   ; f=0.080541
+        +el: name=Carbon     ; f=0.599846
+        +el: name=Sulfur     ; f=0.319613
```
SimonRit, line 768: I would chose only one PMMA

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
+        +el: name=Carbon     ; f=0.640
+        +el: name=Nitrogen   ; f=0.008
+        +el: name=Oxygen     ; f=0.229
+        +el: name=Phosphor   ; f=0.002
+        +el: name=Calcium    ; f=0.001
+        +el: name=Scandium   ; f=0.0
+        +el: name=Titanium   ; f=0.0
+        +el: name=Vandium    ; f=0.0
+        +el: name=Chromium   ; f=0.0
+        +el: name=Manganese  ; f=0.0
```
SimonRit: I would remove, I don't think there is one Tumor material

GateMaterials.db
```diff
@@ -476,16 +664,241 @@ Testis: d=1.04 g/cm3 ; n=9
...
+        +mat: name=FR4       ; f=0.76
+
+FR4_Copper_epoxy: d=2.2998 g/cm3; n=7; state=solid
+        +el: name=Gold       ; f=0.0131
+        +el: name=Nickel     ; f=0.0182
+        +el: name=Copper     ; f=0.2134
+        +el: name=Carbon     ; f=0.4848
+        +el: name=Hydrogen   ; f=0.0486
+        +el: name=Oxygen     ; f=0.1538
+        +el: name=Chlorine   ; f=0.0681
```
SimonRit, line 829: Any clue what's FR4?

SimonRit (commented):
Thanks for the huge work. A couple of inline comments. It would be helpful to sort materials in alphabetical order.

GateMaterials.db
andiresch: Yes, and I would go for PMMA2 as defined in NIST:
https://physics.nist.gov/cgi-bin/Star/compos.pl?refer=ap&matno=223

GateMaterials.db
djboersma: According to French wikipedia, Hostaphan (with ph instead of p) one of the brand names under which a foil of "Polytéréphtalate d'éthylène" (a.k.a. PET) is known. Another name for exactly the same material is Mylar. I think the name Mylar is much more common, so I am keeping Mylar and deleting "Hostapan".

GateMaterials.db
djboersma: Thanks

GateMaterials.db
djboersma: I agree, let's stick to English.

GateMaterials.db
djboersma: Keeping A

GateMaterials.db
djboersma: Removing all definitions of Air*

GateMaterials.db
djboersma: removing all defintions of Air*

GateMaterials.db
djboersma: done

GateMaterials.db
djboersma: kept only PMMA2 (and renamed it to PMMA)

GateMaterials.db
djboersma: removed

GateMaterials.db
djboersma: removing both `Adipose` and `Adipose_Tissue` in favor of `G4_ADIPOSE_TISSUE_ICRP`

GateMaterials.db
djboersma: agreed

GateMaterials.db
djboersma: https://en.wikipedia.org/wiki/FR-4

looks like a material typically used for circuit boards. maybe important for simulation of SiPMs?

i think i will keep these materials for now

GateMaterials.db
djboersma: agreed

GateMaterials.db
djboersma: I could not find any information about what "VLC" means in this context. But in all references that I looked at while searching the interwebs about lanthaum bromide, the density was always 5.06 or 5.08 g/cm3, I could not find any that has 5.29. So I am keeping only one LaBr3 material and I drop the VLC suffix.
https://en.wikipedia.org/wiki/Lanthanum%28III%29_bromide

PR #398, opened 2021-01-13
add LMF and ECAT to sources/externals

djboersma:
Instead of underhandedly distributing unmaintained source tar balls and instructing the users to follow cumbersome build procedures, put that code in "externals" and run (a variation on) those compilation procedures via cmake in case ECAT and/or LMF are selected in the cmake step.

I'm doing this pull request already now so that we can discuss it in tomorrow's technical board meeting.
Please do not merge until:
* the code has been tested
* compile warnings are somehow silenced
* a decision is taken about whether to keep the old method for compiling and linking with these packages, or to rip that out completely (e.g. if the `ECAT7_HOME` environment variable is set, try to use it, otherwise build the source tree under `source/externals/ecat`) 
* docs for build & install are also updated (should be an update to this same PR)
* copyright issues are cleared


dsarrut:
Still work in progress. @djboersma said : 

"About #398: I made those additional commits in an attempt to make it work also on my shiny new MacOSX machine, while travelling without my rusty old linux laptop. The clang compiler has different sensitivities than gcc/g++. In LMF, it needed an extra line in CMakeLists.txt to allow for modern C++ language elements in ROOT 6.22.08 (the current production version, but they are probably present in all 6.* versions). In ECAT there are bigger problems. The 'makefilters.c' function used locally defined/declared functions *before* they were defined/declared. So I changed the order of the functions in the source file, and that seemed to solve it, only to reveal a more serious problem: the code uses a function called "xfrf_", which is completely undefined as far as can see. At least it's not defined in the ECAT sources. In other projects I have seen the underscore suffix as a convention for linking to Fortran libraries. But I don't see any indication that this is indeed happening. I think then it should also be #including the "cfortran.h" header file, which is not happening. So I am stumped. I am able to compile the code on Linux with gcc/g++ 9.3.0 (LinuxMint 21), and when I run "nm" on the compiled binary, grep shows me lots of functions with "ecat" in it, but none with xfrf. Maybe the optimizer removes the code that uses xfrf before the linker discovers that it does not exist? While clang is less tolerant of illegal code?

I just recompiled the lmf_ecat branch on my Linux laptop. The compilation is noisy (I did not invest any time into silencing the warnings, that's the job of a real maintainer), but it does indeed still compile. I also ran the (one and only) PET ECAT example from GateContrib. It seems to run without any problems. So I think that my clang-induced fixes have not broken anything. Note that I do not have a single LMF example macro, so I do not know if the LMF functionality can indeed still be run without crashing.

* I suspect/hope that merging my PR can still be useful for LMF users on both Linux and MacOSX. But I had no opportunity at all to really test it.
* I am even more hopeful that it is useful for ECAT users on Linux.
* For ECAT users on MacOSX, merging the PR is counterproductive. If you do that, they should be advised to use the older Gate releases and compile ECAT themselves in some way."

djboersma:
Thanks David! :)

PR #402, opened 2021-01-20
Try to debug brew on github actions

tbaudier:


djboersma:
I think that this just illustrates why we should try to get rid of ROOT in Gate. At the moment we basically only use ROOT for storing output: phase spaces (`TTree`), images and dose distributions (`TH2`, `TH3`). Maybe we should enable the conversion utilities for phase spaces and images in GateTools to also write ROOT output, so that users who for some reason really need their data in ROOT format can use that to do that to still get their work done.

PR #405, opened 2021-01-25
updates and improvements in LET- and EnergySpectrum-Actor

andiresch:
LET Actor:
- Scoring of LET to other material, which can be specified by the user
- W value in air can be calculated from two parametrizations
- A maximum and minimum LET threshold can be set by the user

EnergySpectrum Actor:
- Scoring of LET to other material, which can be specified by the user
- Log binning of histograms now available to all histograms
- improved code readability 
- changed output format - now identical to TEPCActor output, namely center of bin instead of lower edge (first column); second column: bin width (this is new) and third column frequency (which was 2nd column before)


dsarrut:
Thanks a lot Andreas!

Issue #399, opened 2021-01-14
imageCT output not based on energy


DavidLeibold:
I noticed that the imageCT output for imaging applications does not output an image based on the total energy collected in one pixel, which is what I would have expected based on the name. I checked that instead, the imageCT values represent the number of singles per pixel. I am not sure whether this is a bug or a feature. If it is the latter, I guess an edit in the respective [section of the documentation](https://opengate.readthedocs.io/en/latest/data_output_management.html#image-ct-output) would be helpful to point this out.

dsarrut:
Hi, thank for pointing out. Could you please modify the documentation ? (Edit on github in the small menu at bottom right). Thanks !

DavidLeibold:
I modified the documentation and created a pull request.

dsarrut:
perfect, thanks!

PR #412, opened 2021-01-30
Update data_output_management.rst

DavidLeibold:
Adding description of how the values of the image CT output are calculated, as discussed in issue #399.


dsarrut:
thanks!

PR #416, opened 2021-02-18
In GateEnergySpectrumActor.cc, modify edep to edepEvent to fill the histogram

tbaudier:
With edep > 0, it was always false because edep is equal to 0 and the histogram is always empty

(cc @narbor)


andiresch:
yes that is true. Thank you for spotting this bug. With charged particle beams you wouldn't see this problem, because edep should almost never be 0 in that case.

PR #418, opened 2021-03-04
Fixed: twice saved trees to root output file

Crauzerus:
**Problem:**
An output generated by GateToRoot class generates doubled trees.
doubled_trees_gate_to_root

**Solution:**
Set an overwrite option enable by replacing _Write()_ with _Write(0,TObject::kOverwrite)_

This solution solves described problem for all output channels implemented in the GateToRoot.hh file.


dsarrut:
really useful thanks !

PR #421, opened 2021-03-11
enable decay with all PL

dsarrut:


tbaudier:
@dsarrut  This commit introduced an error in the benchmarks for test 3 (range e+). The output (range.root) with this commit is empty

dsarrut:
good catch! Here is a PR: https://github.com/OpenGATE/Gate/pull/425

PR #423, opened 2021-03-13
Add custom electromagnetic field maps

dhaberl:
Hello, 

I added custom electromagnetic field maps similar to GateElectricTabulatedField3D.cc and GateMagTabulatedField3D.cc.

The electromagnetic field is generated from from a look-up table (e.g., txt file) containing the spatial positions and their associated electric and magnetic field strengths.

The field can be applied using the GATE command:
/gate/geometry/setElectMagTabulateField3D

Best, 
David


dsarrut:
great! Could you please
- update the doc to include your changes (https://opengate.readthedocs.io/en/latest)
- provide an example of macro that tests this feature, ideally with a simple python script that checks the results? We can include it in the benchmark list (https://github.com/OpenGATE/GateBenchmarks)

(You are working with Hermann I suppose?)

thanks!
David

dhaberl:
Yes, I am working with Hermann.

Sure, I will provide an example for testing and update the documentation including the electric and magnetic field since they are also not documented yet.

dhaberl:
I updated the doc with my changes.

I will add the documentation of the existing GateElectricTabulatedField3D.cc and GateMagTabulatedField3D.cc in a separate pull request, because I would like to propose some changes there.

dsarrut:
thanks!

PR #424, opened 2021-03-29
set cxx14 (for recent pytorch)

dsarrut:
For recent libtorch (>1.7), I had to specify this flag.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -243,3 +243,45 @@ jobs:
           ls -lrt $GITHUB_WORKSPACE/benchmarks/benchRT/output
         fi
 
+  benchmarks:
+    runs-on: ubuntu-latest
+    needs: [build]
+    strategy:
+      fail-fast: false
+      matrix:
+        strategy_name: [t1_edep_profiles, t2_edep_vox, t3_range_e+, t4_necr, t5_pet, t6_dpk, t7_garf, t9_gaga_phsp]
+
+    steps:
+    - name: Checkout github repo
+      uses: actions/checkout@v2
+      with:
+          lfs: true
+    - name: Checkout LFS objects
+      run: git lfs checkout
+    - name: Checkout submodules
+      run: git submodule update --init --recursive
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install -y gcc \
+                                g++ \
+                                make \
+                                binutils \
+                                libx11-dev \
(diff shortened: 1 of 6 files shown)


SimonRit:
Don't you need to set this with the same value as Geant4? In Gate, I would set it accordingly and you could raise an error if this is not the correct value for Pytorch.

dsarrut:
I don't know ... I think G4 do not need this flag, only required for Gate when compiled with libtorch> 1.7.

SimonRit:
Looking at `UseGeant4.cmake` and `Geant4Config.cmake` in my Geant4 build directory, this is already set in Gate by Geant4. @dsarrut, have you compiled Geant4 with cxx11?

dsarrut:
I cannot find it in Geant4, may be it depends on the system?

SimonRit:
I'm not sure what you can't find. I was referring to [this](https://github.com/Geant4/geant4/blob/master/cmake/Templates/Geant4Config.cmake.in#L194) and [this](https://github.com/Geant4/geant4/blob/master/cmake/Templates/UseGeant4.cmake#L47). From what I understand (which is not much), they don't use `CMAKE_CXX_STANDARD` (like you) but `COMPILE_FEATURES` [here](https://github.com/Geant4/geant4/blob/master/cmake/Modules/G4BuildSettings.cmake#L202). But I think the result is the same, setting a flag to use the selected c++ standard.

> (and no, I did not set any cxx11 in Gate compilation)

I meant did you compile Geant4 with the same cxx14 flag? On my computer, the default is ` GEANT4_BUILD_CXXSTD=11`.  Gate (without your patch) will compile with the same flag as Geant4 by default. I wonder if you can compile gate with c++14 and Geant4 with c++11.

dsarrut:
no, I did not recompile Geant4 with cxx14, only Gate. Currently, Gate works when linked to the default compiled Geant4 (so I assume cxx11?). 
It seems to me that Thomas suggestion could be nice, only put this flag when using libtorch. I don't know if specifying this cxx may lead to issues...

SimonRit:
Some elements [here](https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects). Not very clear to me, are we passing `std::string` to Geant4 from Gate? I find it less risky to enforce the same in Gate and Geant4 (as it is currently on master I think, but you could check). But I understand it's annoying to have to recompile Geant4.

BTW, does torch accept c++17? Because your patch won't allow it (anymore).

PR #430, opened 2021-04-08
revert commit 2e20a49fe9621fb1fa7f6332add99d89cf75e90d

wrzof:
commit 2e20a49fe9621fb1fa7f6332add99d89cf75e90d added global 

set (CMAKE_CXX_STANDARD 14)

which was unwanted.


dsarrut:
Thanks Mathieu ! (there is still an issue with test7, but I will check that ...)

PR #432, opened 2021-04-10
remove old benchmarks

djboersma:
The old benchmarks are now obsolete, the benchmark tests are now developed and maintained in a new GateBenchMarks repository. One of the weaknesses of the old tests was that they relied on data.kitware.com for the input data. Another weakness was that they relied on exact equality of simulation results to reference data. The new tests are statistics based, like they should, for a Monte Carlo program!

I kept the CTest line in CMakeLists.txt, so that the "make test" command still runs and returns 0. Otherwise the CI stuff might break, I suspect. Maybe TB can correct/comment this.

I hope that at some point we will actually merge the unit test branch into develop. Then we will need CTest again.


djboersma:
I see that the CI build fails anyway. I see that in `.github/workflows/main.yml` there are some lines that seem to do an `ls` on the old benchmarks output folders, which don't exist anymore. Can we just comment that out? Maybe much more from that main.yml file should be cleaned up, now that the "benchmarks" folder goes away?

tbaudier:
Thank you @djboersma , I removed all mentions to the old test in ```.github/workflows/main.yml```
I did that test to check in Github actions that a simple simulation is running correctly on all tested platforms without testing the result.
We can consider that the new benchmarks are more efficient for that but with the new benchmarks, we do not try the tests on MacOS

djboersma:
Thanks @tbaudier , looks like this makes the CI test successful. But it worries me that you write that with the new benchmarks, you are not testing MacOS. I think that is an important target to test. Because there are MacOS users, and also because by testing different platforms you'll catch different errors. So I hope that at last some of the new benchmarks at some point can also be run with MacOS.

tbaudier:
I'm agree with you, tests on MacOs are important. For the moment, I cannot run the benchmarks on MacOS because I used a docker image (on Ubuntu). To be able to run them on MacOs, we need to adapt it.

PR #442, opened 2021-06-03
Update issue templates

tbaudier:
Add template for bug report


m-dupont (changes requested):
I think we can also ask for the compiler version. And for GATE, the version or the hash of the commit.

tbaudier:
Thank you for your comments

PR #451, opened 2021-07-16
Coincidences by event identification option  and additional info about primary photon interactions

anetxe:
Information of current number of Compton, Rayleigh and Pair Creation interactions of primary photons (parentID=0) in sensitive detectors  is included
An option to sort singles into coincidences by event identification is added. This is particularly useful for vpgTLE source that does not include time information.


dsarrut:
thanks !

PR #465, opened 2021-10-12
Update defining_a_geometry.rst

DavidLeibold:
The table "Commands of the sub-tree geometry for different shapes", e.g. box, sphere etc., is not shown in the html version, maybe because the vertical lines of a couple of cells did not align?


dsarrut:
thanks !

PR #477, opened 2022-01-04
Try to debug with new version of MacOS

tbaudier:
Change the cache to update root and geant4 compilation


tbaudier:
@cpommranz @djboersma @dsarrut 
I force to use c++17 during Gate compilation. I used https://stackoverflow.com/a/33525982 and https://stackoverflow.com/a/48978022

Are you're ok with that solution?

PR #479, opened 2022-01-07
Small modifications of Digitizer

kochebina:
1) Bug in the definition of a hit with max deposited energy (for Adder with takeEnergyWinner policy)
2) Go to next event in a HitCollection is empty (i.e. no hits to process)
3) Classes GatePulseProcessorChain, GatePulseProcessorChainMessenger, GatePulseProcessorMessenger and GateVPulseProcessor are moved from physics to digits_hits


kochebina:
Oups. I've lost Gate.cc somewhere on the way. Add it now. Sorry about that

Issue #480, opened 2022-01-10
Valgrind detects use of uninitialized values


maxaehle:
Summary
=======
I ran `valgrind` on GATE v9.1 and got a lot of *Conditional jump or move depends on uninitialised value(s)* messages. I checked the first and fourth one and believe that valgrind is correct there, see below for more details. The complete log is attached here:  [valg2-1.txt](https://github.com/OpenGATE/Gate/files/7838588/valg2-1.txt) [valg2-2.txt](https://github.com/OpenGATE/Gate/files/7838593/valg2-2.txt) [valg2-3.txt](https://github.com/OpenGATE/Gate/files/7838601/valg2-3.txt) (split into three files due to file size limits).

If the warnings are actually correct, GATE could behave non-deterministically even if the random number generator's seed is fixed. Similar issues are known in ROOT, but the developers seem to have investigated the warnings there and provide a [suppression file](https://github.com/root-project/root/blob/e1e22198f55729f46f81f5274022e287e405ce57/etc/valgrind-root.supp) for valgrind. I used this suppression file, so the findings are caused by Gate/Geant4.

**Has this been investigated? If yes, with which result?**

Details regarding the first finding
===================================
```
==59== Conditional jump or move depends on uninitialised value(s)
==59==    at 0xD1F54A: GateCrystalSD::PrepareCreatorAttachment(GateVVolume*) (GateCrystalSD.cc:245)
==59==    by 0x11FF5D9: GateVVolume::AttachCrystalSD() (GateVVolume.cc:388)
==59==    by 0x1209FA5: GateVolumeMessenger::SetNewValue(G4UIcommand*, G4String) (GateVolumeMessenger.cc:83)
==59==    by 0x104CB659: G4UIcommand::DoIt(G4String) (G4UIcommand.cc:262)
==59==    by 0x104E06E4: G4UImanager::ApplyCommand(char const*) (G4UImanager.cc:584)
==59==    by 0x104DFE71: G4UImanager::ApplyCommand(G4String const&) (G4UImanager.cc:481)
==59==    by 0x104C6561: G4UIbatch::ExecCommand(G4String const&) (G4UIbatch.cc:182)
==59==    by 0x104C6790: G4UIbatch::SessionStart() (G4UIbatch.cc:230)
==59==    by 0x104DEAFD: G4UImanager::ExecuteMacroFile(char const*) (G4UImanager.cc:308)
[...]

PR #483, opened 2022-01-27
Fix uninitialized value, first finding in #480

maxaehle:
Initialize [`GateCrystalSD::m_systemList`](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/include/GateCrystalSD.hh#L86) with `nullptr`.

Currently the member is not initialized. I believe `nullptr` is the right initial value because of [this check](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/src/GateCrystalSD.cc#L245).

This fixes a *Conditional jump or move depends on uninitialized value(s)* message by valgrind, the first finding mentioned in #480. To verify that, I ran valgrind for Gate v9.1 (with a configuration that I could probably share if requested)
[without](https://github.com/OpenGATE/Gate/files/7949618/valg-withsupp-orig.txt) and [with](https://github.com/OpenGATE/Gate/files/7949620/valg-withsupp-change1.txt) the proposed change, using the [suppression file](https://github.com/root-project/root/blob/e1e22198f55729f46f81f5274022e287e405ce57/etc/valgrind-root.supp) of the ROOT project. I already removed the PIDs in the attached logs, and compared them with `diff`. The first message in #480 is gone, and apart from that some pointers changed, which I think is irrelevant.

### DIFF
--- a/source/digits_hits/include/GateCrystalSD.hh
+++ b/source/digits_hits/include/GateCrystalSD.hh
@@ -83,7 +83,7 @@ class GateCrystalSD : public G4VSensitiveDetector
 
   protected:
      GateVSystem* m_system;                           //! System to which the SD is attached //mhadi_obso obsollete, because we use the multi-system approach
-     GateSystemList* m_systemList;                    //! System list instead of one system
+     GateSystemList* m_systemList = nullptr;          //! System list instead of one system
   private:
       GateCrystalHitsCollection * crystalCollection;  //! Hit collection
 


maxaehle:
> I believe `nullptr` is the right initial value because of this [check](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/src/GateCrystalSD.cc#L245).

To expand on this, as a test, I initialized `GateCrystalSD::m_systemList` with `static_cast<GateSystemList*>(nullptr)+1`(=0x18) instead of `nullptr`(=0x0). This is just a random value that this uninitialized member could happen to have. With this change, Gate segfaults in the second if statement of the following snippet ([from `GateCrystalSD::PrepareCreatorAttachment`](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/src/GateCrystalSD.cc#L252)):
```c++
   if (!m_systemList)
   {
      m_systemList = new GateSystemList;
      AddSystem(creatorSystem);
   }
   else
   {
      if((m_systemList->back()) != creatorSystem)
         AddSystem(creatorSystem);
   }
```
This piece of code assumes that if `m_systemList` is not a `nullptr`, it is a pointer to a `GateSystemList` (so we can call its function `back`). Gate segfaulted because with the above change, `m_systemList` was neither `nullptr` nor pointing to any allocated memory.

In the other PR #484 I'm not sure what value the uninitialized variable should be initialized with.

Issue #471, opened 2021-12-07
ignoring return value in GateMergeManager.cc:346


olahaye74:
Fail to build GateMergeManager using -Wunused-result

Compiling GateMergeManager.cc...
./src/GateMergeManager.cc: In member function 'void GateMergeManager::StartCleaning(std::string, bool)':
./src/GateMergeManager.cc:346:12: warning: ignoring return value of 'int system(const char*)', declared with attribute warn_unused_result [-Wunused-result]
      system(rmfiles.c_str());
      ~~~~~~^~~~~~~~~~~~~~~~~

dsarrut:
Done #487

PR #487, opened 2022-02-15
Solved issue 471

Crauzerus:
From my tests I didn't receive an error mentioned in the [issue 471](https://github.com/OpenGATE/Gate/issues/471)
I checked it with two versions of g++ (9.3.0 and 7.1.0).

However, the reason why this error can raise is caused by calling system(...) without catching returned value.
What it is interesting is fact that there are unnecessary lines of code - an author of code executes two commands:

`rm -f dir/*`

`rm -f -r dir`

which can be done with one command:

`rm -rf dir`

Because a returned values is only required to check its output (0 - command successed, 1 - command failed) there is no point to assign this output to the variable - instead, it is better to redirect this to if(...).


dsarrut:
Thanks Mateusz !

PR #488, opened 2022-02-17
Fix uninitialized value, two more findings

maxaehle:
## Second finding in #480 
```
Conditional jump or move depends on uninitialised value(s)
    at 0x10F3D1A: GateHitTree::Init(GateRootHitBuffer&) (GateRootDefs.cc:286)
    by 0xFB8FDE: GateToRoot::Book() (GateToRoot.cc:225)
    ...
```
is due to the uninitialized variable `GateAnalysis::m_recordSeptalFlag`, whose value is read [here](https://github.com/OpenGATE/Gate/blob/01a705adb0aa059402b6fab3682fb80c0e08d9d6/source/general/src/GateRootDefs.cc#L74) and returned to decide on [this](https://github.com/OpenGATE/Gate/blob/01a705adb0aa059402b6fab3682fb80c0e08d9d6/source/general/src/GateRootDefs.cc#L286) if-statement. 

To verify that it is not initialized, I grepped for the variable name and found that it is only used four times: In the above if-statement, in the declaration, and in the getter and setter functions in [`GateAnalysis.hh`](https://github.com/OpenGATE/Gate/blob/01a705adb0aa059402b6fab3682fb80c0e08d9d6/source/digits_hits/include/GateAnalysis.hh). I set a breakpoint to the setter function which was never reached in the setup of #480.

`m_recordSeptalFlag` seems to control whether some data should be collected and outputted. Its default should be `false`.

## Third finding in #480 
The third finding 
```
Conditional jump or move depends on uninitialised value(s)
    at 0xFBEED5: GateToRoot::RecordOpticalData(G4Event const*) (GateToRoot.cc:919)
    by 0xFBE715: GateToRoot::RecordEndOfEvent(G4Event const*) (GateToRoot.cc:808)
```
is due to the uninitialized variable `GateToRoot::m_rootOpticalFlag`. According to grep, the variable name occurs in four contexts: the [declaration](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/include/GateToRoot.hh#L340), [getter](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/include/GateToRoot.hh#L250) and [setter](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/include/GateToRoot.hh#L252) in `GateToRoot.hh`, and [here](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/src/GateToRoot.cc#L919) in an `if` condition in `GateToRoot.cc`. With gdb I found out that the getter and setter are never called, but the `if` condition is evaluated several times. Again with grep I found that the setter is only called from a single location [here](https://github.com/OpenGATE/Gate/blob/86083caae3a4f41903b3d1e3c3d5571e807e041e/source/digits_hits/src/GateToRootMessenger.cc#L139) in `GateToRootMessenger::SetNewValue`.

`m_rootOpticalFlag` also seems to control whether some data is collected and outputted.  Its default should be `false`.

This fix also solves the subsequent zlib related findings such as
```
Use of uninitialised value of size 8
   at 0x10A27B2D: deflate_fast (deflate.c:1849)
   by 0x10A2601A: deflate (deflate.c:1002)
   ...
```

Using vgdb in different setups I verified that the zlib related errors occur precisely when `m_rootOpticalFlag` happens to be `true` . 

## Remark
Running the pct-bug-geometry setup of #480 in valgrind again after fixing the four uninitialized variables (`m_recordSeptalFlag` and `m_rootOpticalFlag` in this PR, `m_systemList` in #483, `mparticle_time` in #484) gives no *Uninitialized variable* or *Invalid read* messages any more. Other setups that we are currently investigating still produce such messages which we will investigate next, so let's keep #480 open.

## Acknowledgements
@ggonei @NIoSaT


ggonei:
To summarise the problem and solution, from a [discussion](https://project-gitlab.ztt.hs-worms.de/pct/pct-bug-geometry/-/issues/1) with the above users:
- uninitialised flags mean the flags can sometimes be set to true incorrectly and run different parts of the code
- the different parts of the code happen to either directly call, or be related to, Fill() events
- these Fill() events write data to branches which are uninitialised also, which is expected as the relevant flags relate to processes which require more information in the initial mac file to actually create valid data
- as an example, `m_rootOpticalFlag` would only be set to true by the user if the user had also included information about `OpticalPhoton`s, thus running the relevant `OpticalPhoton` methods to generate the correct data for the tree
- The `OpticalPhoton` code should be investigated further for those users who are interested, as e.g. the lack of a phantom could cause a crash even with other relevant information set up

I requested that Max check initialisation of the buffer and that seems to contain all the key event and hit information, which looks fine.  In general, it may be worth a more in depth look at the initilisation of trees, and checking that the variables utilised by branches are initialise with some clearly 'correct' but 'wrong' information (e.g. -1 for # of hits).  Perhaps other data is being written which is deemed valid (in the sense it does not cause a crash), but is not detected by Valgrind (because Valgrind does not see anything wrong with a reference to an unitiliased variable (&variable) since the reference is simply a memory address and does exist), but is still uninitialised and so contains garbage.  If that garbage happens to be valid (and for a branch of `signed double`, it will always be valid as a double can contain anywhere from `00 00 00 00` -> `FF FF FF FF`) we may not see an error but we may have non-deterministic data (as the memory addresses pointed to are not guaranteed to contain the same bits).  In fact, this error is only observed from a fill of a 'C' branch, and likely is an issue with null bytes within strings.  Even changing the path of where the program runs can change the crash behaviour, as previously null bytes are written with different information.

If this is relevant I suggest opening a new issue which references this one.  At present, a warning message should be given to the user when setting `/gate/output/root/setRootOpticalFlag 1` in the macro without any of the relevant optical photon requirements, as otherwise the crash will be seen again.

PR #490, opened 2022-02-22
Remove additional hash symbol in txt tree output

schillingalex:
There is an additional '#' at the beginning of the unified tree output in txt format. This makes it incompatible with other output formats, because the 'PDGEncoding' column in other formats is called '#PDGEncoding' in txt.

If there is a use for the '#', I would be very interested to hear about it.

### DIFF
--- a/source/io/src/GateAsciiFile.cc
+++ b/source/io/src/GateAsciiFile.cc
@@ -124,7 +124,6 @@ void GateOutputAsciiTreeFile::write_header()
     if( (m_mode & ios_base::out) != ios_base::out )
         throw GateModeFileException("OutputAsciiTreeFile::write_header: file not opened in write mode"); // Should not happen
 
-    m_file << "#";
     uint32_t i = 0;
     for (auto&& d : m_vector_of_pointer_to_data) // access by const reference
     {


dsarrut:
Hi, I think the first use of '#' was to allow to parse commented lines. I am unsure if it is still useful or not here (I admit I don't use this ascii output). If we remove this '#', it will break compatibility. Any advice, @wrzof ?

m-dupont:
For me, it's quite standard to have # on not relevant (comments) lines. Like in gate dose actor for example. 

> This makes it incompatible with other output formats

Why ? Because of the lack of space after '#' ?

schillingalex:
> Why ? Because of the lack of space after '#' ?

If you interpret the txt output as CSV (e.g. with Python's Pandas library), the first column will be called "#PDGEncoding". Combining this with other runs where the output format may have been root or npy will not work, because the column is named "PDGEncoding".

I guess this is why the format is called txt and not CSV. If a change to this breaks existing implementations, it should of course not be included and I will be doing some special handling of that case. I understand that compatibility between different formats is a very niche application.

dsarrut:
yes, let's keep compatibility for the moment.

PR #495, opened 2022-03-12
Set cxx standard, reconfigure root

SimonRit:
I have been struggling to compile the develop branch of gate due to issues with cxx standard of root. I have now implemented a more meaningful message but that enforces the same cxx standard or Geant4, Gate and root. If you like it, it would be great to check that it works before merging, in particular the torch part. Thanks!


tbaudier:
The modifications seems good for me.
I just want to adapt the Github actions workflow:
https://github.com/OpenGATE/Gate/blob/develop/.github/workflows/main.yml#L80-L166

If I understand well:
 - root need c++17 flag
 - Geant4 v11 does not need c++17 flag anymore
 - Gate does not need c++17 flag anymore
 - libtorch need at least c++14
 - RTK need c++17 flag ?

I will try to push the modification on you branch to see the behavior
Thank you

SimonRit:
> root need c++17 flag

My suggestion is to force the user to use the same as Geant4 (so c++17, c++20 or c++23, see below)


> Geant4 v11 does not need c++17 flag anymore

Geant4 only allows 17, 20 and 23 (if available):
https://gitlab.cern.ch/geant4/geant4/-/blob/master/cmake/Modules/G4BuildSettings.cmake#L164-L168

> Gate does not need c++17 flag anymore

Idem as root, I set it equal to Geant4's. 

> libtorch need at least c++14

I don't know if one can use different C++ versions....

> RTK need c++17 flag ?

It compiles with c++14 but I haven't tested it yet...

PR #497, opened 2022-03-15
Fix source collimation when setting the geometry from an RTK file

SimonRit:
See
http://lists.opengatecollaboration.org/pipermail/gate-users/2022-March/012872.html


GabrieleBelotti:
Fix was tested and solved the issue mentioned in the mailing list.
I compared results for primary and secondary images produced with an .xml and an equivalent .placements file, both with source and panel offsets for multiple angles.

SimonRit:
Thanks for the feedback @GabrieleBelotti! I'll let @dsarrut merge...

tbaudier:
Thank you !

PR #499, opened 2022-03-18
Add nicer error message for unknown property when xml file is read. 

m-dupont:
Check availability of property in G4MaterialPropertiesTable before calling AddProperty (avoid exception and backtrace) 
And show message to help the switch from G4.10 to G4-V11.


dsarrut:
Indeed, nicer for user ;) 
I wait for the tests to complete before merging. 
thanks !

Issue #155, opened 2017-11-20
Deletion of GateMultiSensitiveDetector::pMultiFunctionalDetector and the Segfaults


Supernabla:
Deleting a `G4MultiFunctionalDetector` invokes the deletion of the registered primitives. Consequently, the deletion of `GateMultiSensitiveDetector::pMultiFunctionalDetector` invokes the deletion of Gate actors. There are two scenarios in which `GateMultiSensitiveDetector::pMultiFunctionalDetector` is deleted:

1. `delete pMultiFunctionalDetector` is called in `GateMultiSensitiveDetector::~GateMultiSensitiveDetector()`
2. `GateMultiSensitiveDetector::pMultiFunctionalDetector` is registered to the `G4SDManager` which takes the ownership.

People have been commenting in and out these cases in there commits, e.g. bda725bdccd9ade5d12f343c70464b3de241471f, because they figured it causes segfaults.

The deeper problem is the following: the singleton `GateActorManager` also claims ownership of all Gate actors, implementing their deletion in `GateActorManager::~GateActorManager()`. That causes a double free, if `GateMultiSensitiveDetector::pMultiFunctionalDetector` is deleted, too.

dsarrut:
You're right sorry ... What's the solution in your opinion ? Remove the loop that delete all actor in GateActorManager::~GateActorManager() ?

Supernabla:
First of all: thanks for the quick reply and thanks for the work you put into Gate.

Yeah, what you suggest is the provisional fix I was thinking of: Allow the deletion of `GateMultiSensitiveDetector::pMultiFunctionalDetector` and remove the loop in `GateActorManager::~GateActorManager()`.

On the long term, the `GateVActor` <--> `GateActorManager` <--> `GateMultiSensitiveDetector` interfaces might need some more invasive refactoring.

dsarrut:
ok, done 12dd4c4f4267d10cda6ee6bed70ac869c903b28e

Supernabla:
Maybe, I was too quick with my suggestion... @dsarrut I saw, that you encountered problems again (see 8a23dd500ea70ba226f7dc6d9af04611210b23be), so maybe this issue should be re-opened.

Two important points:
* At the moment, no actors are deleted, neither by `GateMultiSensitiveDetector::pMultiFunctionalDetector` nor by the `GateActorManager`, right?

[...]

PR #503, opened 2022-03-25
Delete actors at the end of the simulation

m-dupont:
Revert 12dd4c4f4267d10cda6ee6bed70ac869c903b28e This commit was already a revert (due to issue #155). In this commit, the destructor only delete actors which are not given to MultiFonctionalDetctor.


dsarrut:
thanks Mathieu !

PR #504, opened 2022-03-29
Update LET Actor user manual

andiresch:
- added three possible scoring types:  Method 'A' from Cortes-Giraldo 2014, Average kinetic Energy and the g_QQ0 correction factors for EBT3 films in proton beams; Those methods were available before, but not documented
- removed the documentation of restricted LET calculation as the implementation is not validated
- added explanation for parallel computation


dsarrut:
Thanks!

PR #508, opened 2022-04-04
Fix gjs compilation against Geant4 v11.0

### LINKED ISSUE #507: gjs fails to compile in GATE v9.2
Albert described on the Gate-users mailing list, that gjs fails to compile with the GATE v9.2 release (http://lists.opengatecollaboration.org/pipermail/gate-users/2022-March/012931.html).

This is due to the breaking changes to the G4String interface introduced with Geant4 v11.0.0.

cpommranz:
closes #507


dsarrut:
great thanks ! (I dont use gjs so I was not aware) ; maybe a benchmarks will be needed ...

PR #538, opened 2022-06-24
Check for parallel geometry only if source is a plane.

### LINKED ISSUE #351: RTK Error in SPECT_FFD Example
Hello,

I am having trouble running the SPECT_FFD example in GateContrib. When I run the example that is provided with vGate V9.0 without any modification, it gives the following error:
GateFixedForcedDetectionActor.cc (l.1575): RTK requires equal source-to-center and center-to-detector distances in parallel geometry

I have tried changing various translation and placement values but the error has not gone away and I am not familiar enough with RTK to know what else to try, especially as the error is present with the original example.

Cheers,
Miriam

SimonRit:
Parallel geometry is not a problem if the source is isotropic.
Fixes #351


SimonRit:
@arobert01

Issue #553, opened 2022-11-18
Visualization in DOCKER


ignacio-scarinci:
Hello, the docker version does not allow to use of the visualization. I was looking for a solution to this problem, and I was able to solve it by simply adding the following line in the DockerFileGeant file:
mesa-dri-drivers

```
RUN yum install -y make
                    libX11-devel.x86_64 \
                    libXpm-devel.x86_64 \
                    libXft-devel.x86_64 \
                    libXext-devel.x86_64 \
                    libXmu-devel.x86_64 \
                    libxml2-devel.x86_64 \
                    binutils \
                    qt5-qtbase-devel.x86_64 \
                    git \
                    cmake \
                    wget \
                    mesa-libGLU-devel.x86_64 \ 
                    python3-devel \
                    python3-devel \
                    which nano \
                    mesa-dri-drivers
```

I think that in the following versions, this could be added to be able to have the possibility to use visualization.

Regards

BishopWolf:
Please, add the changes yourself. You know this is a collaborative work! ;)

dsarrut:
Thanks, Ignacio, we will have a look!

BishopWolf:
The PR has been merged. This topic may be closed @dsarrut . 
Thanks @ignacio-scarinci

dsarrut:
thanks !

PR #554, opened 2022-11-21
Adding visualization in docker solution

BishopWolf:
Solution proposed in https://github.com/OpenGATE/Gate/issues/553 by @ignacio-scarinci

### DIFF
--- a/source/docker/DockerFileGeant
+++ b/source/docker/DockerFileGeant
@@ -24,7 +24,8 @@ RUN yum install -y  make \
                     wget \
                     mesa-libGLU-devel.x86_64 \
                     python3-devel \
-                    which nano
+                    which nano \
+                    mesa-dri-drivers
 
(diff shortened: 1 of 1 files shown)


BishopWolf:
@dsarrut @tbaudier  you shall rebuild all docker images with this change, so all images can display visualization

dsarrut:
thanks a lot Alex! We will try this in the following weeks, thanks

ignacio-scarinci:
Hi, 

The proposed solution works flawlessly on Linux.

However, on macOS, there is a bug in xQuartz (https://github.com/XQuartz/XQuartz/issues/144) which is easily fixed with the following command:

`defaults write org.xquartz.X11 enable_iglx -bool YES`

tbaudier:
Thank you, I merged. The next Docker images will integrate the modification.

PR #565, opened 2023-01-21
Update GateMiscFunctions.cc to prevent premature movements when using placements lists

pmcorreia:
When using movements with the list of placements in a text file, for a given time the geometry is now updated according to the highest preceding (past) time index defined in the list rather than the closest time (index), which may lead to (premature) movements occur earlier than the desired (list declared) time.

Question raised here https://listserv.in2p3.fr/cgi-bin/wa?A2=ind2212&L=OPENGATE-DEV-L&X=C617C8E29EB97C0CB9&Y=pmcorreia%40ua.pt&P=1302


dsarrut:
thanks a lot !

PR #570, opened 2023-02-01
Adding another link to getting started with ROOT

MaxTousss:
A user in the mailing proposed a document that is made for GATE user. I believe it should either be linked or hosted with the GATE documentation since ROOT is one of the main output format. Also, the previous link seems to require a login. It might be the same document, but a link without additional login seems better.


tbaudier:
Thank you very much, I merged

PR #580, opened 2023-03-10
Update GateMDBCreators.cc

tbaudier:
Change element->GetNumberOfElements() to element->GetNumberOfIsotopes()
@narbor


mbalcerzyk:
The error is not fixed in https://github.com/OpenGATE/Gate/blob/develop/source/geometry/src/GateMDBCreators.cc, so the normal compilation of Gate 9.4 and docker image "latest" keep failing for isotopes.

Issue #615, opened 2023-06-01
vGate9.3 maybe with memory leak


Fantasticlyy:
**Describe the bug**
use the vGate 9.3,while use digitizer and enable the “/gate/output/” ，when run a simulation ,the RAM will be raise by follow time,I thin may be  enable the “/gate/output/” with memory leak,this issue only appear in vGate9.3 ,the other version not appear.
**Desktop (please complete the following information):**
 vGate9.3 VirtualBox 7.06

#############
# DIGITIZER #
#############
/gate/digitizerMgr/name LESingles
/gate/digitizerMgr/chooseSD pixel
/gate/digitizerMgr/insert SinglesDigitizer
/gate/digitizerMgr/pixel/SinglesDigitizer/LESingles/setInputCollection Singles
/gate/digitizerMgr/pixel/SinglesDigitizer/LESingles/insert energyFraming
/gate/digitizerMgr/pixel/SinglesDigitizer/LESingles/energyFraming/setMin 0. keV
/gate/digitizerMgr/pixel/SinglesDigitizer/LESingles/energyFraming/setMax 6000. keV

########
# OUTPUT #
########
/gate/output/tree/enable
/gate/output/tree/addFileName output/LHCsI.txt
/gate/output/tree/hits/enable               
/gate/output/tree/addCollection LESingles
/gate/output/tree/addCollection Singles

tontyoutoure:
Can confirm. 
I ran valgrind and the following part is what increases as duration of simulation increase:
<details>
  <summary>parts of valgrind output</summary>

```
==72618== 2,844,528 bytes in 118,522 blocks are definitely lost in loss record 22,855 of 22,861
==72618==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==72618==    by 0x359D45: __gnu_cxx::new_allocator<int>::allocate(unsigned long, void const*) (new_allocator.h:114)
==72618==    by 0x358F84: std::allocator_traits<std::allocator<int> >::allocate(std::allocator<int>&, unsigned long) (alloc_traits.h:443)
==72618==    by 0x3584AD: std::_Vector_base<int, std::allocator<int> >::_M_allocate(unsigned long) (stl_vector.h:343)
==72618==    by 0x37E712: std::_Vector_base<int, std::allocator<int> >::_M_create_storage(unsigned long) (stl_vector.h:358)
[...]

PR #619, opened 2023-06-03
fix memory leak in digitizer, issue #615

tontyoutoure:


tontyoutoure:
Major part of the leak is fixed, still some leak. Will check more carefully.

Edit: turns out to be my fix about resolution digitizer. This fix for main branch is sufficient.

kochebina:
Thank you @tontyoutoure !

PR #622, opened 2023-06-08
fix: make GateSpatialResolution do CheckIfEnoughLevelsAreDefined only once

tontyoutoure:


granadogmarc (commented):
This issue has been addressed in a more recent commit.

Issue #639, opened 2023-10-20
 Attach actor to "world" volume causing summary to malfuntion.


tontyoutoure:
**Describe the bug**
Attach actor to "world" causing summary to malfuntion.
1. hits and singles are written twice in the summary file
2. singles count is doubled

**Desktop (please complete the following information):**
 - OS: ubuntu 20.04
 - Gate version or commit hash: commit 07f892ca
 - Geant4 version: 11.1.1
 - Root version:  v6.26/08.
 - compiler version:  gcc 9.4.0

**Minimal example**
Pretty easy to replicate, just attach an actor to world.

**Expected behavior**
that summary works fine

**Additional context**
I will try to fix it and submit a PR.

tontyoutoure:
after a little digging it seems to me that the problem is around `GateActorManager.cc:304`. The comment there said:

_// Remove attached SD by replacing with a deactivated clone SD_

However, it clearly did not remove the attached sd. Instead, every hit is processed twice, once by the original CrystalSD, once by the mutifuctional SD.

also, I checked the volume's sd before and after "replacing", `volume->GetLogicalVolume()->GetSensitiveDetector()` remains the same address. 

edit: Hits are not processed twice. Just singles count are doubled in the summary.

tontyoutoure:
At the time of PR #354 , this seems to be a smart solution. Now it has become the issue itself. Hopefully @cpommranz and @dsarrut can look into this. I'm happy to provide a barebone replication of this issue by then. Without your people's help, I will have to dig deeper to be comfortable enough to provide a fix.

tontyoutoure:
Yes, there is also memory leak. The hero seems to turn into dragon now.

dsarrut:
ouch ;) Thanks for your investigation! We should look at this. However, to be fair, we are committed to opengate version 10 right now (python version), and have very little bandwidth to dig into this., so it may take a bit of time.

tontyoutoure:
@dsarrut Yes I suppose, looking forward to the python ver! I will see if I can fix it myself. However some (desynchronized) Q/A will be much appreciated.

dsarrut:
Sur, I can try to help no problem!

tontyoutoure:
Well, after a afternoon of digging, I confirmed that, when one tries to tie an actor to a volume that contains a CrystalSD, the following will happen:
[...]

PR #640, opened 2023-10-25
fix memory leak and summary output bug

tontyoutoure:
Fix problems described in issue #639.


tontyoutoure:
Add a new one without creating new files.

PR #646, opened 2023-11-27
Fixed forced detection fixes

SimonRit:


SimonRit:
@arobert01 Would that also fix FFD SPECT issues? How can we test this?

PR #665, opened 2024-02-03
Add mixmaxrng

cpommranz:
Add MixMaxRng as an option to the list of random engines. MixMaxRng is the default Geant4 RNG since version 10.4. Note that the default GATE RNG (JamesRandom) is not changed.


dsarrut:
Thanks ! Note this is the default in Gate10

PR #674, opened 2024-03-28
Increase required RTK version to 2.5 for accurate fixed forced detection

SimonRit:
Commit ddc188532c306d464ee9ce9989f3d17b7be97e34 accounts for a required RTK change, RTKConsortium/RTK@a4ba257d39c2c1f1b7eef79f59631cc7bb2005ae, which has been released starting with RTK v2.5.


SimonRit:
My bad, it's actually there since version 2.0 so no need to increase the required RTK apparently

PR #691, opened 2024-07-24
Monolithic crystal

granadogmarc:
Adding of a new digitizer GateVirtualSegmentationSD. The Virtual Segmentation Digitizer module provides a mechanism to generate an ID based on a virtual segmentation of the monolithic crystal, aligned with its spatial resolution. Ideally, the pitch size should be at least half of the position resolution. This virtual segmentation occurs post-simulation, ensuring that the simulation speed remains uncompromised even when dealing with large systems and numerous crystals such as total-body PET scans.


tontyoutoure:
Hi, may I ask why you need a virtual segmentation. Shouldn't some simple python scripting with np.histogram2d of the hit position be sufficient?

granadogmarc:
Hi Li Kun,

first of all, sorry for the late reply, I was on holiday when you sent your
email and this was deep down in my inbox and it took a while to get to it.

The idea behind this was to make it available for other people working with
monolithic crystals; and avoid the extra handling of big data files. We aim
to work with a total-body PET scanner and the amount of data will be
enormous. Thus, we thought this would be an optimised way to deal with this
issue.

Cheers,
M

PR #727, opened 2025-03-05
Position resolution

granadogmarc:
Do CheckIfEnoughLevelsAreDefined only once as requested in the Pull Request with ID #622


kochebina:
@granadogmarc Hi Marc, it seems there is an issue with this PR during the GATE compil.
Could you check it, please?

PR #744, opened 2026-01-22
Default digitizer memleak fix

graag:
When we switched from GATE 9.0 to GATE 9.4 I noticed severe memory leaks. The slower mem leak was fixed by #732.

The larger mem leak is caused by the fact that GATE allows to define a new digitizer with the same name as the default `Singles` one. After reading the documentation on the new digitizer we came up with following definition:

```
/gate/digitizerMgr/chooseSD crystal_lso

/gate/digitizerMgr/name     Singles
/gate/digitizerMgr/insert   SinglesDigitizer
```

Such definition leads two digit collections being created both named `Singles_crystal_lso`. Both collections will gather all digits in an event. However as they are stored by name only one of them will be registered in an event for cleanup. This leads to a severe mem leak.

Screenshot_20250918_080325 (1)

It can be worked around by not defining the digitizer and using the default one or by using a different name for the digitizer. However it was not obvious to us at the beginning.

Proposed changes stop simulation with an error message if user tries to define a digitizer with the same name as the default one. This PR is based on top of #743. All changes relevant to this PR are in the last commit.


graag:
I'm not sure about this t27_hits_2SD test. I've seen the "Wasserstein distance test for posY failed" error when running benchmarks locally. However re-running it gives good result:

```
Entries test is OK:  7679  vs.  7474
Mean edep test is OK:  0.23610407  vs.  0.22799999
Median edep test is OK:  0.21403126  vs.  0.20340845
Mean posX test is OK:  -0.8977521  vs.  -3.3175309
Median posX test is OK:  -0.9064665  vs.  -1.9669635
Mean posY test is OK:  -1.2801986  vs.  5.4492297
Median posY test is OK:  -2.1562946  vs.  12.357294
Mean posZ test is OK:  -0.48724806  vs.  -0.17989987
Median posZ test is OK:  -0.0011477646  vs.  0.0011456421
Mean time test is OK:  0.02487176729123031  vs.  0.024969286961905278
Median time test is OK:  0.024935102249110004  vs.  0.024793800909225717
output/cylPET.root test is OK!
Entries test is OK:  12436  vs.  12385
Mean edep test is OK:  0.18372092  vs.  0.18368773
Median edep test is OK:  0.16575754  vs.  0.16343597
Mean posX test is OK:  0.13432513  vs.  -0.8890376
Median posX test is OK:  -0.0153078055  vs.  -0.5256139
Mean posY test is OK:  8.013136  vs.  7.936506
Median posY test is OK:  7.357194  vs.  7.230829
Mean posZ test is OK:  0.11849357  vs.  0.03490759
Median posZ test is OK:  -5.258529e-05  vs.  -4.9508846e-05
Mean time test is OK:  0.10102452867231385  vs.  0.09976677502044554
Median time test is OK:  0.101864200839301  vs.  0.10044976746419068
output/SPECT.root test is OK!
```

Is this something that should be investigated or is this posY distribution noisy??

PR #746, opened 2026-02-03
Spatial resolution 2D distribution changes

granadogmarc:
Generalisation of the 2D distribution. Adding options to map axis from file to crystal and providing different distributions for each axis.


CMakeLists.txt
kochebina, line 1: Hi Marc, @granadogmarc 
Can you, please, comment what this part is for?

granadogmarc: Ah, my bad! I'll take this out, I used it to run som tests but obviously there is no need to keep it in the Git! Sorry for that! I'll remove it tomorrow if that's alright!

kochebina: @granadogmarc , don't worry, I'll do it ;-)

PR #762, opened 2026-04-16
Correction of the acollinearity sampling following Toussaint et al.

mraedler:
Adapt the correction suggested by Toussaint et al. (https://doi.org/10.1088/1361-6560/ad70f1) to fix the sampling of the of the acollinearity for back-to-back simulations, i.e. when using
/gate/source/\<name\>/setAccolinearityFlag True
/gate/source/\<name\>/setAccoValue 0.5 deg

Also add the acollinearity effect to simulations with positron sources. When using
/gate/source/\<name\>/gps/particle e+
/gate/source/\<name\>/gps/ene/type <Fluor18, Oxygen15, Carbon11>
acollinearity is not simulated. One needs to set the MeanEnergyPerIonPair value from 0 to 5 eV (https://www.geant4.org/download/release-notes/notes-v10.7.0.html). This was also remarked by Toussaint et al. 

This PR  also adds an option to set the MeanEnergyPerIonPair value via
/gate/geometry/setMeanEnergyPerIonPair \<material\> \<value\> \<unit[energy]\>
This is completely analogous to setting the ionization potential value (I-value) of a material. It also changes the default value of the MeanEnergyPerIonPair to 5 eV, if the value is not set.

### DIFF
--- a/source/geometry/include/GateDetectorConstruction.hh
+++ b/source/geometry/include/GateDetectorConstruction.hh
@@ -159,7 +159,8 @@ public:
   void SetMaterialIoniPotential(G4String n,G4double v){theListOfIonisationPotential[n]=v;}
   G4double GetMaterialIoniPotential(G4String n){ return theListOfIonisationPotential[n];}
 
-
+  void SetMaterialMeanEnergyPerIonPair(G4String n,G4double v){theListOfMeanEnergyPerIonPair[n]=v;}
+  G4double GetMaterialMeanEnergyPerIonPair(G4String n){ return theListOfMeanEnergyPerIonPair[n];}
 
 private :
 
@@ -185,6 +186,7 @@ protected :
   G4bool moveFlag;
 
   std::map<G4String,G4double> theListOfIonisationPotential;
+  std::map<G4String,G4double> theListOfMeanEnergyPerIonPair;
 
 
 private:
--- a/source/geometry/include/GateDetectorMessenger.hh
+++ b/source/geometry/include/GateDetectorMessenger.hh
@@ -58,6 +58,7 @@ class GateDetectorMessenger: public G4UImessenger
 
     G4UIcmdWithoutParameter*   pListCreatorsCmd;
     G4UIcmdWithAString*        IoniCmd;
+    G4UIcmdWithAString*        IonPairCmd;
 
     //G4UIcmdWithABool* 	       pEnableAutoUpdateCmd;    
     //G4UIcmdWithABool* 	       pDisableAutoUpdateCmd; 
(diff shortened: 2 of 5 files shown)


MaxTousss:
Greetings!

The current PR implements APA, as suggested in the paper, for BackToBack source. 

However, it will not change anything for positron sources. If this is the intent of the PR than the PR should be good! If not, a decision need to be made about accepting, or not, the "risk" of changing the value of "MeanEnergyPerIonPair" for all materials.

Whichever is chosen, I suggest that a PR to modify the GateBenchmark be made to show that it works and that it would be linked here, with an image of the results, to show that it does indeed work.

Bests,
Maxime

mraedler:
Regarding the benchmark I have a commit to fix it here
https://github.com/mraedler/GateBenchmarks/commit/85d191ba3b71ab7792ee60ae5d2ebe2932142e23

I could open a PR on 
https://github.com/OpenGATE/GateBenchmarks

mraedler:
And here are the plots
The original benchmark fits a Gaussian
acoDistComparison_gauss_fit

The fixed version fits a Rayleigh distribution
acoDistComparison_rayleigh_fit

mraedler:
Regarding the positron sources:
I suppose (correct me if i'm wrong) we would need to insert

material->GetIonisation()->SetMeanEnergyPerIonPair(5. * eV);

in /source/geometry/src/GateMaterialDatabase.cc below line 94

That would probably activate it in all cases. One could think of adding an option to turn it off.

@MaxTousss let me know what you think and btw thank you for commenting!

MaxTousss:
That is indeed the command suggested by Geant4 to activate APA as seen in PET. However, to be truthful, I am not sure where it should be defined. For the simulations made for the article, I placed it in 
`source/geometry/src/GateDetectorConstruction.cc` and it worked. I guess that the materials that can be used in GateDetector are also those used everywhere else...

In particular, this is where I put it:
```  
const G4MaterialTable * theTable = G4Material::GetMaterialTable();
  for(unsigned int i =0;i<(*theTable).size();i++){
    (*theTable)[i]->GetIonisation()->SetMeanEnergyPerIonPair(5.0*eV);     <---- The command
    if(theListOfIonisationPotential[(*theTable)[i]->GetName()]){`
(...)
```  
Your suggestion of placement seems more general but I can't say if it is better, worst or incorrect. Sorry!

As for an option to deactivate it, I do not have a strong opinion on that. Note that you could also give an option to change the value. The shape and width of APA change with temperature, so, maybe someone somewhere would like to play with that value with compiling each time.

Again, I repeat that my knowledge of GATE is too limited to know if this command change anything else than APA. Thus, I leave GATE dev to decide on that. 

Finally, it is a pleasure to be able to help!

